PR #14, opened 2022-11-16
Introduce compose BoM from October 22' release

### LINKED ISSUE #8: Update to Compose October '22
Would be great to update the Compose samples to use the Compose BOM. See https://github.com/android/user-interface-samples/pull/399 for an example of how to do such a migration.

oas004:
## Why?

Resolves #8 

## How?

In this PR I have updated to use the new Bill of Material (BoM) from the October release (https://developer.android.com/jetpack/androidx/releases/compose#2022.10.00). In order to do this I had to update Kotlin to 1.7.20 and update the compose compiler version to 1.3.1. I also had to bump the target and compile versions for the compose modules. I noticed that the other modules also has a quite low compile/target versions. Is this on purpose? I can also bump this if that is desirable.

Furthermore, would it be nice to introduce a libs.toml file to keep track of the dependency versions? I would be happy to contribute if that is something we want :)


compose/recomposehighlighter/build.gradle
```diff
@@ -35,18 +35,21 @@ android {
...
     }
     composeOptions {
-        kotlinCompilerExtensionVersion compose_version
+        kotlinCompilerExtensionVersion compose_compiler_version
     }
 
 }
 
 dependencies {
+    def composeBom = platform('androidx.compose:compose-bom:2022.10.00')
```
JolandaVerhoef: Can you update this to `2022.11.00`?

JolandaVerhoef (approved):
LGTM, only needs the latest greatest BOM release :)

compose/recomposehighlighter/build.gradle
oas004: Good point! Changed it in 2a9dff2 :)

PR #25, opened 2022-12-08
Add apply spotless workflow.

arriolac:
Automatically runs `./gradlew spotlessApply` on a new PR and pushes any updates to the same PR.

Example of it running: https://github.com/arriolac/snippets/pull/2

### DIFF
--- a/.github/workflows/apply_spotless.yml
+++ b/.github/workflows/apply_spotless.yml
@@ -11,18 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-name: Build snippets
+name: Apply Spotless
 
 on:
-  push:
-    branches: [ '*' ]
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/apply_spotless.yml
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+name: Apply Spotless
+
+on:
+  pull_request:
+    branches: [ '*' ]
```
bentrengrove: Not that it matters, but I'm pretty sure this line is redundant

bentrengrove (approved):
Very cool

.github/workflows/apply_spotless.yml
arriolac: Yeah it looks like it still runs without it. Thanks!

PR #28, opened 2023-01-06
Fixing Brush: Gradients and Shaders snippet duplication

astamato:
As per https://developer.android.com/jetpack/compose/graphics/draw/brush#image-as-brush there is a duplication in the code snippet for usage of brush in Text.


riggaroo:
thanks @astamato

PR #74, opened 2023-02-27
Added lazy layout snippets

riggaroo:


d16066240-hue:
compose/snippets/src/main/java/com/example/compose/snippets/lists/LazyListSnippets.kt

hamen:
## Round 5 dispositions (reviewed SHA adbde19, fixed in f6ec103)

| Finding | From | Classification | Outcome |
|---|---|---|---|
| Long form accepts a placeholder in the other comment syntax | codex | Qualifying: correctness | Fixed. The placeholder must match the marker's syntax. Tests for both mismatches. |
| D22, D16 and a code comment omit "that hides code" | opencode 1 | Qualifying: correctness | Fixed in all three. |
| Review skill offers the private wrapper to registered components | grok 1 | Qualifying: correctness | Fixed. |
| Extraction skill reads minSdk from the catalog only | grok 2 | Qualifying: correctness | Fixed. |
| Skills name only `// ...` for the long form | grok nit | Nit | Fixed. |
| One placeholder between two silent excludes counts twice | opencode nit 2 | Not qualifying | Rare shape; recorded for a follow-up. |
| Verdict says "elision only" when a group also has a commented import | opencode nit 3 | Not qualifying | The report's import section lists it. Recorded. |
| `handWrittenOf` undercount; `mapping-long.json` has no `lang` | grok nits | Not qualifying | Recorded. Informational list only; null-language pairing is the documented permissive case. |

415 tests pass. #1082: old silent form exits 1, plain form exits 0.

PR #81, opened 2023-03-02
Add snippets for Pager

riggaroo:


riggaroo:
@arriolac can you double check this PR? I've had to change one import for material 3 bottom sheets, when updating to Compose 1.4.0

compose/snippets/build.gradle
```diff
@@ -50,46 +50,46 @@ android {
 dependencies {
 
     implementation 'androidx.paging:paging-common-ktx:3.1.1'
-    def composeBom = platform('androidx.compose:compose-bom:2023.01.00')
```
arriolac, line 53: Nit: Should we continue to use the BOM and override versions as needed?

compose/snippets/src/main/java/com/example/compose/snippets/layouts/MaterialLayoutSnippets.kt
```diff
@@ -375,7 +375,7 @@ fun BottomSheetDemo() {
     // [END android_compose_layout_material_bottom_sheet]
 
     // [START android_compose_layout_material_bottom_sheet2]
-    val sheetState = rememberSheetState()
+    val sheetState = rememberModalBottomSheetState()
```
arriolac, line 378: The CL for the corresponding DAC page still needs review from a Material SME, but `rememerModalBottomSheetState` should be the correct one. Thanks.

arriolac (approved):

PR #89, opened 2023-04-19
Updated Android gradle tools to v7.4.2 and gradle wrapper to v7.6 and others

lawgimenez:
I have noticed several outdated dependencies while checking it out and decided to update them.


riggaroo:
Thanks for the ping here, we've updated the versions in a bigger refactor.

PR #112, opened 2023-06-07
Make custom column's children not restricted by column's height

hrach:
This fixes the layout and make the example fit the description:

> The child composables are constrained by the Layout constraints (without the minHeight constraints), and they're placed based on the yPosition of the previous composable.


simona-anomis:
@hrach thanks for your PR. Could you please provide more info why you think this is needed? Since it's a simple code snippet, we just want the incoming parent constraints to be used as they are in this case

simona-anomis:
@hrach ah I see. My interpretation was that, since it's a simple code snippet,  "without the minHeight constraints" means we won't change any constraints, not even minHeight. But your interpretation is opposite, more matching the expected Column behaviour.

In that case, could you please move this to line 158 (you've made a change in the wrong snippet) and leave a comment above, explaining why you're overriding the child items constraints? Thanks 😊

PR #113, opened 2023-06-14
Migrate build scripts to kts, and version catalogs.

### LINKED ISSUE #19: Update tests so that they pass
So that we can run the `build` workflow on the entire project. 

Context: #18

riggaroo:
Migrate all build.gradle files to kts.
Move to version catalogs for dependency management.
Delete slices module. 

Fixes #19


build.gradle.kts
```diff
@@ -0,0 +1,8 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+plugins {
+    alias(libs.plugins.gradle.versions)
+    alias(libs.plugins.version.catalog.update)
```
zsmb13, line 4: We should list all the plugins that the subprojects use here, with `apply false` added so that they're loaded but not applied at the root, e.g.

```
    alias(libs.plugins.android.application) apply false
```

settings.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+        mavenCentral()
+    }
+}
+rootProject.name = "snippets"
+include(
+        ":shared",
+        ":bluetoothle",
+        ":compose:recomposehighlighter",
+        ":kotlin",
+        ":compose:snippets"
```
zsmb13: Nit:

```suggestion
        ":compose:snippets",
```

shared/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+            isMinifyEnabled = false
+            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"),
+                    "proguard-rules.pro")
+        }
+    }
+}
+dependencies {
+    implementation(fileTree("libs") {
+        include("*.jar")
+    })
```
zsmb13: This is hopefully not required

riggaroo: Good point, removed.

settings.gradle.kts
riggaroo: Done

build.gradle.kts
riggaroo: Thanks, didnt get why i needed this.. Done now :D

zsmb13 (approved):
LGTM!

calren (approved):

PR #130, opened 2023-07-26
Navigation

jakeroseman:
Adds a simple landing screen to the app that allows the user to choose which snippets they would like to view. So far I've just added the two pre-existing screens that host the snippets for images and brushes. It should be easy to extend this with new screens where we want to demonstrate certain snippets.


compose/snippets/src/main/java/com/example/compose/snippets/LandingScreen/LandingScreen.kt
```diff
@@ -0,0 +1,78 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+@Composable
+fun LandingScreen(
+    toBrushExamples: () -> Unit,
```
riggaroo: What do you think about making something like a Destination enum that contains a list of destinations, with their paths, then we can put them in a LazyColumn list of items.

I did this in my playground of experiments and it works pretty well when adding more and more snippets. https://github.com/riggaroo/compose-playtime/blob/main/app/src/main/java/dev/riggaroo/composeplaytime/MainActivity.kt#L56

jakeroseman: Done. I mimicked your approach and implemented it in this [commit](https://github.com/android/snippets/pull/130).

compose/snippets/src/main/java/com/example/compose/snippets/SnippetsActivity.kt
```diff
@@ -19,31 +19,45 @@ package com.example.compose.snippets
...
-                    ) {
-                        // BrushExamplesScreen()
-                        ImageExamplesScreen()
+                    NavHost(navController, startDestination = "LandingScreen") {
+                        composable("LandingScreen") {
+                            LandingScreen(
+                                toBrushExamples = { navController.navigate("BrushExamples") },
+                                toImageExamples = { navController.navigate("ImageExamples") },
+                            )
+                        }
```
riggaroo: This could be removed

compose/snippets/src/main/java/com/example/compose/snippets/LandingScreen/LandingScreen.kt
riggaroo: Nice, you can remove the specific callbacks here then, and just have a generic, `onNavigateTo: (Destination) -> Unit` parameter, then you only need one callback.

compose/snippets/src/main/java/com/example/compose/snippets/LandingScreen/LandingScreen.kt
```diff
@@ -0,0 +1,90 @@
...
+fun NavigationButtons(toBrushExamples: () -> Unit, toImageExamples: () -> Unit) {
+    LazyColumn(
+        modifier = Modifier
+            .padding(16.dp)
+            .fillMaxSize(),
+        verticalArrangement = Arrangement.spacedBy(8.dp),
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+        items(Destination.values().toList()) { destination ->
+            when (destination) {
```
riggaroo: As well as this `when` would be simplified to just one `NavigationButton()` (i think this should also be called `NavigationItem`), that just passes the destination to the callback.

compose/snippets/src/main/java/com/example/compose/snippets/SnippetsActivity.kt
jakeroseman: Done.

compose/snippets/src/main/java/com/example/compose/snippets/LandingScreen/LandingScreen.kt
jakeroseman: Done. I more thoroughly copied the example you linked earlier.

compose/snippets/src/main/java/com/example/compose/snippets/LandingScreen/LandingScreen.kt
jakeroseman: I'm not sure if I've understood correctly what you mean by having just one `NavigationItem`. How does it look now?

riggaroo: You should be able to remove the `when` and just have 
```
NavigationItem(destination) {
                    navigate(
                        destination
                    )
                }
```
Since the branches of the when are doing the same thing.

jakeroseman: Got it! Nice spot. Pushed in [this commit](https://github.com/android/snippets/pull/130/commits/de74866cf5b11716d8e1102f36e03b5e07f54cc0).

riggaroo:
Passed here https://github.com/android/snippets/actions/runs/5669457829/job/15362202406

compose/snippets/src/main/java/com/example/compose/snippets/LandingScreen/LandingScreen.kt
Djqubert, line 1: Screenshot_20240313_001620_Google

PR #134, opened 2023-08-11
Add variable font snippets

riggaroo:
variable_font_custom_axis
variable_fonts


compose/snippets/src/main/java/com/example/compose/snippets/text/VariableFontsSnippets.kt
```diff
@@ -0,0 +1,511 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.example.compose.snippets.R
+
+private object VariableFontSnippet1 {
+    // [START android_compose_text_variable_fonts_define]
+    // In Typography.kt
+    @OptIn(ExperimentalTextApi::class)
+    val DisplayLargeFontFamily =
```
simona-anomis: Nit: should this be capitalized? If so, we should probably change it [in the docs](https://developer.android.com/jetpack/compose/text/fonts#use-downloadable-fonts) as well

simona-anomis (approved):

PR #141, opened 2023-08-18
Sample showing how to take a screenshot from a composable

riggaroo:
Adding a sample that shows how to draw a composable into a bitmap, using `Picture` and `drawIntoCanvas`.


compose/snippets/src/main/java/com/example/compose/snippets/graphics/AdvancedGraphicsSnippets.kt
```diff
@@ -0,0 +1,265 @@
...
+        )
+        Text(
+            "Into the Ocean depths",
+            fontSize = 18.sp
+        )
+    }
+}
+
+fun createBitmapFromPicture(picture: Picture): Bitmap {
+    val pictureDrawable = PictureDrawable(picture)
```
nadewad: IIUC it looks like this PictureDrawable is used only to query the intrinsic width/height. Instead I think you can just query Picture#getWidth/Picture#getHeight instead.

riggaroo: Thanks! removed.

compose/snippets/src/main/java/com/example/compose/snippets/graphics/AdvancedGraphicsSnippets.kt
```diff
@@ -0,0 +1,267 @@
...
+                .fillMaxSize()
+                .drawWithCache {
+                    // Example that shows how to redirect rendering to an Android Picture and then
+                    // draw the picture into the original destination
+                    val width = this.size.width.toInt()
+                    val height = this.size.height.toInt()
+
+                    onDrawWithContent {
+                        val pictureCanvas =
+                            androidx.compose.ui.graphics.Canvas(
```
JolandaVerhoef, line 154: why is there a whole package definition needed?

JolandaVerhoef (approved):
Love it! One small comment.

compose/snippets/src/main/java/com/example/compose/snippets/graphics/AdvancedGraphicsSnippets.kt
riggaroo: Because we need to use two different kinds of canvas' in the snippet, I'll import the other one too so its clearer in the snippets.

PR #171, opened 2023-11-21
Migrate snippets from Material 3 

riggaroo:


compose/snippets/src/main/java/com/example/compose/snippets/designsystems/Material3Snipppets.kt
```diff
@@ -0,0 +1,478 @@
...
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.BaselineShift
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.example.compose.snippets.R
+import com.example.compose.snippets.ui.theme.Typography
+
+private object Material3Snippets {
+    // [START android_compose_material3_experimental_annotation]
+    // import androidx.compose.material3.ExperimentalMaterial3Api
```
arriolac, line 74: Nit: Might this be confused for commenting this import out? Perhaps phrasing it like "Make sure to import..."?

arriolac (approved):
LGTM just a minor comment about a comment 😆

compose/snippets/src/main/java/com/example/compose/snippets/designsystems/Material3Snipppets.kt
riggaroo: Thanks Chris - we've done it like this in a few other places so i think its okay..

PR #175, opened 2023-11-24
Update ProgressIndicator.kt

sanskarjain2110:
It seems there might be a mistake in the usage of the LinearProgressIndicator widget. Instead of providing a lambda function, the progress argument typically expects a float value. Consider revising the code to pass a float value directly, ensuring accurate and intended progress representation.
>
> @Composable
> @ComposableTarget
> public fun LinearProgressIndicator(
>      **progress: Float,**
>     modifier: Modifier,
>     color: Color,
>     trackColor: Color,
>     strokeCap: StrokeCap
> ): Unit


riggaroo:
Hi! 
The progress indicator lambda is a more performant solution to setting progress, since it skips recomposition phase. 
The lambda option is preferred - For more information on this - see https://issuetracker.google.com/295616656

PR #177, opened 2023-12-29
Add snippets for graphics-shape 

riggaroo:


compose/snippets/src/main/java/com/example/compose/snippets/graphics/ShapesSnippets.kt
```diff
@@ -0,0 +1,329 @@
...
+        layoutDirection: LayoutDirection,
+        density: Density
+    ): Outline {
+        /*val matrixA = calculateMatrix(shapeA.bounds, size.width, size.height)
+        shapeA.transform(matrixA)
+        val matrixB = calculateMatrix(shapeB.bounds, size.width, size.height)
+        shapeB.transform(matrixB)
+        val morph = Morph(shapeA, shapeB)
+        morph.progress = percentage*/
+        // todo figure out how to scale and offset the shapes here
```
riggaroo: @chethaase What would be best to change the scale / offset of the shape without having access to the shape or recreating it here?

IanGClifton (approved):

PR #180, opened 2024-01-03
Update material3-adaptive to 1.0.0-alpha03

alexvanyo:
Updates material3-adaptive to 1.0.0-alpha03.

Currently, this doesn't work on smaller screens: tapping a list item appears to have no effect: https://issuetracker.google.com/issues/316376112


IanGClifton:
LGTM though we may want to add a snippet for the MyItem class now that it has a Saver, but we'll need to update the DAC page shortly after.

alexvanyo:
@IanGClifton this is ready for review now, I added a new section for `MyItem` in preparation for including it in the documentation page update.

IanGClifton (approved):

PR #189, opened 2024-01-23
Add pip code to snippets

MagicalMeghan:
Adding in snippets for PiP compose documentation

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
@@ -280,7 +280,7 @@ private fun buildRemoteAction(
 @RequiresApi(Build.VERSION_CODES.O)
 @Composable
 fun PlayerBroadcastReceiver(player: Player?) {
-    if(!isInPipMode() || player == null) {
+    if (!isInPipMode() || player == null) {
         // Broadcast receiver is only used if app is in PiP mode and player is non null
         return
     }
--- a/compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
@@ -1,0 +1,370 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.compose.snippets.pictureInPicture
+
+import android.app.PendingIntent
+import android.app.PictureInPictureParams
+import android.app.RemoteAction
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.ContextWrapper
+import android.content.Intent
+import android.content.IntentFilter
+import android.graphics.drawable.Icon
+import android.os.Build
+import android.util.Rational
+import androidx.activity.ComponentActivity
+import androidx.annotation.DrawableRes
+import androidx.annotation.RequiresApi
+import androidx.annotation.StringRes
+import androidx.compose.foundation.layout.Column
+import androidx.compose.material3.Button
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.toAndroidRectF
+import androidx.compose.ui.layout.boundsInWindow
+import androidx.compose.ui.layout.onGloballyPositioned
+import androidx.compose.ui.platform.LocalContext
+import androidx.core.app.PictureInPictureModeChangedInfo
+import androidx.core.content.ContextCompat
+import androidx.core.graphics.toRect
+import androidx.core.util.Consumer
+import androidx.media3.common.Player
+import androidx.media3.exoplayer.ExoPlayer
+
+var shouldEnterPipMode by mutableStateOf(false)
+
+// [START region_tag_13]
+// Constants for broadcast receiver
+const val ACTION_BROADCAST_CONTROL = "broadcast_control"
+
+// Intent extra for broadcast controls from Picture-in-Picture mode.
+const val EXTRA_CONTROL_TYPE = "control_type"
+const val EXTRA_CONTROL_PLAY = 1
+const val EXTRA_CONTROL_PAUSE = 2
+const val REQUEST_PLAY = 5
+const val REQUEST_PAUSE = 6
+// [END region_tag_13]
+
+@Composable
+fun PipListenerPreAPI12(shouldEnterPipMode: Boolean) {
+    // [START region_tag_1]
+    // [START region_tag_10]
+    val currentShouldEnterPipMode by rememberUpdatedState(newValue = shouldEnterPipMode)
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O &&
+        Build.VERSION.SDK_INT < Build.VERSION_CODES.S
+    ) {
+        val context = LocalContext.current
+        DisposableEffect(context) {
+            val onUserLeaveBehavior = {
+                if (currentShouldEnterPipMode) {
+                    context.findActivity()
+                        .enterPictureInPictureMode(PictureInPictureParams.Builder().build())
+                }
+            }
+            // [END region_tag_10]
+            context.findActivity().addOnUserLeaveHintListener(
+                onUserLeaveBehavior
+            )
+            onDispose {
+                context.findActivity().removeOnUserLeaveHintListener(
+                    onUserLeaveBehavior
+                )
+            }
+        }
+    }
+    // [END region_tag_1]
+}
+
+@Composable
+fun VideoPlayer(
+    modifier: Modifier = Modifier
+) {
+    // [END region_tag_2]
+    val context = LocalContext.current
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        val pipModifier = modifier.onGloballyPositioned { layoutCoordinates ->
+            val builder = PictureInPictureParams.Builder()
+
+            // Add autoEnterEnabled for versions S and up
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                builder.setAutoEnterEnabled(true)
+            }
+            context.findActivity().setPictureInPictureParams(builder.build())
+        }
+    }
+    // [END region_tag_2]
+}
+
+// [START region_tag_3]
+internal fun Context.findActivity(): ComponentActivity {
+    var context = this
+    while (context is ContextWrapper) {
+        if (context is ComponentActivity) return context
+        context = context.baseContext
+    }
+    throw IllegalStateException("Picture in picture should be called in the context of an Activity")
+}
+// [END region_tag_3]
+
+@Composable
+fun VideoPlayerScreen(
+) {
+    // [START region_tag_3]
+    val context = LocalContext.current
+    Button(onClick = {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            context.findActivity().enterPictureInPictureMode(
+                // the parameters have been set by previous calls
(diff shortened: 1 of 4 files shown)


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 accompanist = "0.32.0"
 androidGradlePlugin = "8.1.2"
-androidx-activity-compose = "1.8.0-rc01"
+androidx-activity-compose = "1.9.0-alpha01"
```
arriolac: What API do you need in `androidx-activity-compose`? Do most snippets require this dependency or just one or two?

If most snippets require it, then I think `latest` is the correct branch to add these snippets to. Otherwise, if most snippets don't need it, then `main` would be a better target branch to merge into.

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+@RequiresApi(Build.VERSION_CODES.O)
+fun listOfRemoteActions(isPlaying: Boolean, context: Context): List<RemoteAction> {
+    return listOf()
+}
+
+@Composable
+fun VideoPlayer4(
+    shouldEnterPipMode: Boolean,
+    modifier: Modifier = Modifier,
+
```
arriolac: Nit: Extra new line

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+) {
+    // [START android_compose_pip_add_remote_actions]
+    val context = LocalContext.current
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
arriolac, line 319: What if the version is less than `O`? Should there be an `else` case?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+        )
+    )
+}
+// [END android_compose_pip_build_remote_action]
+
+// [START android_compose_pip_broadcast_receiver]
+@RequiresApi(Build.VERSION_CODES.O)
+@Composable
+fun BroadcastReceiver(player: Player?) {
+    if (isInPipMode() && player != null) {
```
arriolac: Recommendation: This would be more readable if you do an early return. And perhaps log a warning statement:

```
    if (!isInPipMode() || player == null) {
         // Log statement
    }

    // ...rest of the code
```

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+            PendingIntent.FLAG_IMMUTABLE
+        )
+    )
+}
+// [END android_compose_pip_build_remote_action]
+
+// [START android_compose_pip_broadcast_receiver]
+@RequiresApi(Build.VERSION_CODES.O)
+@Composable
+fun BroadcastReceiver(player: Player?) {
```
arriolac: I think this name can be a bit more descriptive. Maybe `PlayerBroadcastReceiver` or `PlayerControllersReceiver` (I like the latter)?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
arriolac: If >= `O` is required, maybe add a `RequiresApi`?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+            }
+            context.findActivity().setPictureInPictureParams(builder.build())
+        }
+    }
+    // [END android_compose_pip_set_aspect_ratio]
+}
+
+// [START android_compose_pip_build_remote_action]
+@RequiresApi(Build.VERSION_CODES.O)
+private fun buildRemoteAction(
```
arriolac: Not sure I see the value of this function - why not create a `RemoteAction` directly?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+
+@Composable
+fun VideoPlayer2(
+    shouldEnterPipMode: Boolean,
+    modifier: Modifier = Modifier,
+) {
+    // [START android_compose_pip_set_aspect_ratio]
+    val context = LocalContext.current
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
arriolac, line 244: Same comment here - should there be an `else` case?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+        val pipModifier = modifier.onGloballyPositioned { layoutCoordinates ->
```
arriolac, line 248: This modifier doesn't seem to be used anywhere. Is there a composable missing here?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+        val context = LocalContext.current
+        DisposableEffect(context) {
+            val onUserLeaveBehavior: () -> Unit = {
+                if (currentShouldEnterPipMode) {
+                    context.findActivity()
+                        .enterPictureInPictureMode(PictureInPictureParams.Builder().build())
+                }
+            }
+
+            // [END android_compose_pip_pre12_should_enter_pip]
```
arriolac: I think this end tag should be at line 245

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+}
+
+@Composable
+fun VideoPlayer(
+    modifier: Modifier = Modifier
+) {
+    // [START android_compose_pip_builder_auto_enter]
+    val context = LocalContext.current
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        val pipModifier = modifier.onGloballyPositioned { layoutCoordinates ->
```
arriolac, line 72: `modifier` isn't used anywhere. Is that intentional?

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,384 @@
...
+fun VideoPlayerScreen() {
+    // [START android_compose_pip_button_click]
+    val context = LocalContext.current
+    Button(onClick = {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            context.findActivity().enterPictureInPictureMode(
+                // the parameters have been set by previous calls
+                PictureInPictureParams.Builder().build()
+            )
+        }
```
arriolac, line 110: Handle `else` case? Perhaps log a warning statement?

gradle/libs.versions.toml
MagicalMeghan: Most snippets do not need it

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
MagicalMeghan: Oop - surprised spotless didn't catch that

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
MagicalMeghan: ah - RequiresAPI may be a better option here

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,334 @@
...
+    }) {
+        Text(text = "Enter PiP mode!")
+    }
+    // [END android_compose_pip_button_click]
+}
+
+// [START android_compose_pip_is_in_pip_mode]
+@Composable
+fun isInPipMode(): Boolean {
```
arriolac: Just catching this now - I think this needs to be prefixed with remember (i.e. `rememberIsInPipMode`). See guidance here: https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#naming-composable-functions-that-remember--the-objects-they-return

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,334 @@
...
+        Log.i("PiP info", "API does not support PiP")
+    }
+    // [END android_compose_pip_set_aspect_ratio]
+}
+
+// [START android_compose_pip_broadcast_receiver]
+@RequiresApi(Build.VERSION_CODES.O)
+@Composable
+fun PlayerBroadcastReceiver(player: Player?) {
+    if (!isInPipMode() || player == null) {
```
arriolac: Since `isInPipMode` internally `remember`s a mutable state, I think the result should be set to a variable to it's clear that a state read is happening. So:

```
val isInPipMode = rememberIsInPipMode()
if (!isInPipMode || player == null) {
    // ...
} 
```

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
bentrengrove: This doesn't look resolved and without attaching the modifier to anything I don't think it will work

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
bentrengrove: Not resolved, same problem as comment above

gradle/libs.versions.toml
```diff
@@ -34,12 +34,14 @@ maps-compose = "3.1.1"
 material = "1.11.0-beta01"
 material3-adaptive = "1.0.0-alpha04"
 material3-adaptive-navigation-suite = "1.0.0-alpha02"
+media3Common = "1.2.1"
```
bentrengrove: I think media3 shares a version number so you should only have one here and use it twice below

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
bentrengrove: Not resolved, this is a good catch. We have to make sure we follow these guidelines

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,334 @@
...
+        // Broadcast receiver is only used if app is in PiP mode and player is non null
+        return
+    }
+    val context = LocalContext.current
+
+    DisposableEffect(key1 = player) {
```
bentrengrove: nit: `key1 =` isn't required

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
bentrengrove: Not resolved, I think I agree

compose/snippets/src/main/java/com/example/compose/snippets/pictureInPicture/PictureInPictureSnippets.kt
```diff
@@ -0,0 +1,334 @@
...
+        val pipModifier = modifier.onGloballyPositioned { layoutCoordinates ->
+            val builder = PictureInPictureParams.Builder()
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                builder.setAutoEnterEnabled(true)
+            }
+            context.findActivity().setPictureInPictureParams(builder.build())
+        }
+    } else {
+        Log.i("PiP info", "API does not support PiP")
```
bentrengrove: nit: Tag should be extracted to a just a const called `private const val PIP_TAG = "PiP info"` seeing as it is used all through this file

bentrengrove (changes requested):
Looks like the work from Chris's original review still needs to be done

bentrengrove (approved):
LGTM thanks for fixing that up

PR #191, opened 2024-01-24
Add snippets for Indication and Ripple migration

arriolac:
Merging this against `latest` branch which uses the latest Compose versions so that `main` continues to use the latest stable version.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/designsystems/Material2Snippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/designsystems/Material2Snippets.kt
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-@file:Suppress("unused")
+@file:Suppress("unused", "DEPRECATION_ERROR")
 
 package com.example.compose.snippets.designsystems
 
--- a/compose/snippets/src/main/java/com/example/compose/snippets/layouts/PagerSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/layouts/PagerSnippets.kt
@@ -426,7 +426,7 @@ private fun CustomSnapDistance() {
         HorizontalPager(
             state = pagerState,
             pageSize = PageSize.Fixed(200.dp),
-            beyondBoundsPageCount = 10,
+            outOfBoundsPageCount = 10,
             flingBehavior = fling
         ) {
             PagerSampleItem(page = it)
--- a/compose/snippets/src/main/java/com/example/compose/snippets/touchinput/focus/FocusSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/touchinput/focus/FocusSnippets.kt
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("DEPRECATION_ERROR")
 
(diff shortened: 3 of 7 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/designsystems/Material2Snippets.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-@file:Suppress("unused")
+@file:Suppress("unused", "DEPRECATION_ERROR")
```
florina-muntenescu, line 17: if something was deprecated here, i think we should check what it is and try to fix the DAC page.

it's ok to add this for now, but we should at least create a bug internally

compose/snippets/src/main/java/com/example/compose/snippets/layouts/PagerSnippets.kt
```diff
@@ -426,7 +426,7 @@ private fun CustomSnapDistance() {
         HorizontalPager(
             state = pagerState,
             pageSize = PageSize.Fixed(200.dp),
-            beyondBoundsPageCount = 10,
+            outOfBoundsPageCount = 10,
```
florina-muntenescu, line 429: is this change reflected in the guidance as well?

compose/snippets/src/main/java/com/example/compose/snippets/touchinput/focus/FocusSnippets.kt
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("DEPRECATION_ERROR")
```
florina-muntenescu, line 17: same as above

compose/snippets/src/main/java/com/example/compose/snippets/layouts/PagerSnippets.kt
arriolac: This change won't be affected in the guidance on DAC since it's on the `latest` branch (DAC mostly pulls snippets on the `main` branch unless explicitly specified to pull from another branch). However, we'll need to update this snippet once 1.7.0 is stable (we'll have to do that on the `main` branch)

compose/snippets/src/main/java/com/example/compose/snippets/designsystems/Material2Snippets.kt
arriolac: I don't think we want to update the DAC page's snippet yet since 1.7.0 is still in alpha. We'll want to do that once 1.7.0 is stable (and the change would have to do be done on `main`)

compose/snippets/src/main/java/com/example/compose/snippets/touchinput/focus/FocusSnippets.kt
arriolac: Same comment here. I think there may be other snippets we'll have to update once 1.7.0 goes stable on the `main` branch (like if we deprecate other APIs). 

Perhaps it'd be helpful for me to mention our branch policy in the README so that it is clear?

arriolac:
Going to merge this in as it it currently blocking a DAC page release. If there are any comments / concerns I can address that separately.

Note that I also created #192 to keep `main` and `latest` in sync

PR #197, opened 2024-01-30
Migrate recomposeHighlighter to Modifier.Node

mlykotom:


compose/recomposehighlighter/src/main/java/com/example/android/compose/recomposehighlighter/RecomposeHighlighter.kt
```diff
@@ -16,100 +16,143 @@
...
+    /**
+     * The value of totalCompositions at the last timeout.
+     */
+    private var totalCompositionsAtLastTimeout = 0L
+
+    fun incrementCompositions() {
+        totalCompositions++
+    }
+
+    override fun onAttach() {
```
bentrengrove, line 85: Do you need to cancel the timer on detach?

compose/recomposehighlighter/src/main/java/com/example/android/compose/recomposehighlighter/RecomposeHighlighter.kt
```diff
@@ -16,100 +16,143 @@
...
+        debugInspectorInfo { name = "recomposeHighlighter" }
+    }
+
+    override fun create(): RecomposeHighlighterModifier = RecomposeHighlighterModifier()
+
+    override fun update(node: RecomposeHighlighterModifier) {
+        node.incrementCompositions()
+    }
+
+    // It's never equal, so that every recomposition triggers the update function.
```
bentrengrove, line 60: Smart 👍

compose/recomposehighlighter/src/main/java/com/example/android/compose/recomposehighlighter/RecomposeHighlighter.kt
```diff
@@ -16,100 +16,143 @@
...
+            if (field == value) return
+            restartTimer()
+            field = value
+            invalidateDraw()
+        }
+
+    /**
+     * The value of totalCompositions at the last timeout.
+     */
+    private var totalCompositionsAtLastTimeout = 0L
```
bentrengrove: nit: To simplify, I don't think you even need this anymore because we don't do anything with the total over time. I think in the previous version they were using it as a way to trigger redraw, but in your version you have a better option in `invalidateDraw()`. You could just reset `totalCompositions` to `0` on the end of the timer and this would all work.

compose/recomposehighlighter/src/main/java/com/example/android/compose/recomposehighlighter/RecomposeHighlighter.kt
mlykotom: Added 👍🏼

bentrengrove (approved):

PR #205, opened 2024-02-06
Update PiP package name

MagicalMeghan:
Updating PiP package name to not use camel case and updating a code comment


arriolac (approved):
LGTM

PR #210, opened 2024-02-07
Add new Compose text snippets

bentrengrove:


compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
```diff
@@ -577,7 +591,188 @@ private object TextEffectiveStateManagement2 {
     // [END android_compose_text_state_management]
 }
 
-private val firaSansFamily = FontFamily()
+@Composable
+private fun TextSample(samples: Map<String, @Composable ()->Unit>) {
```
arriolac, line 595: This composable used in the snippets below - did you want to include a region tag for these as well?

arriolac (approved):
LGTM just one comment about whether or not `TextSample` should be included

compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
bentrengrove: Left it out deliberately. The screenshots in the doc include labels to show which text paragraph is which because its not super obvious, but then the code snippet wouldn't match the screenshot. I made this as kind of a middle ground where people could click into the snippet to see how it was generated but the code snippet in DAC was still just the relevant code.

PR #228, opened 2024-02-27
Add `DeviceConfigurationOverride` snippets

alexvanyo:
Adds `DeviceConfigurationOverride` snippets using the new 1.7.x test API.


gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ androidGradlePlugin = "8.2.2"
 androidx-activity-compose = "1.9.0-alpha03"
 androidx-appcompat = "1.6.1"
 androidx-compose-bom = "2024.02.01"
+androidx-compose-ui-test = "1.7.0-alpha03"
```
JoseAlcerreca: how about adding a maintenance comment here to remove this version when the bom contains 1.7.0?

JoseAlcerreca (approved):

gradle/libs.versions.toml
alexvanyo: Added!

PR #249, opened 2024-04-11
Update GlanceSnippets.kt

secondsun:
Minor fix for glance docs code.


secondsun:
@riggaroo This should fix most of the problems the twitter user identified with the Glance example. Would you take a look and approve?

riggaroo (approved):

PR #255, opened 2024-04-25
Updated ListDetailPaneScaffold use to alpha11

IanGClifton:
This eliminates storing state outside and directly uses the navigator as the source of truth. This also gets rid of the Modifier for AnimatedPane as it's no longer a required argument.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/adaptivelayouts/SampleListDetailPaneScaffold.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/adaptivelayouts/SampleListDetailPaneScaffold.kt
@@ -16,6 +16,7 @@
 
 package com.example.compose.snippets.adaptivelayouts
 
+import android.os.Parcelable
 import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
@@ -42,6 +43,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import kotlinx.parcelize.Parcelize
 
 @OptIn(ExperimentalMaterial3AdaptiveApi::class)
 @Composable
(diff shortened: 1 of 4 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/adaptivelayouts/SampleListDetailPaneScaffold.kt
```diff
@@ -49,19 +47,13 @@ import androidx.compose.ui.unit.sp
...
     // [START android_compose_adaptivelayouts_sample_list_detail_pane_scaffold_part02]
-    val navigator = rememberListDetailPaneScaffoldNavigator<Nothing>()
+    val navigator = rememberListDetailPaneScaffoldNavigator<MyItem>()
 
     BackHandler(navigator.canNavigateBack()) {
         navigator.navigateBack()
     }
     // [END android_compose_adaptivelayouts_sample_list_detail_pane_scaffold_part02]
 
-    // [START android_compose_adaptivelayouts_sample_list_detail_pane_scaffold_part01]
```
alexvanyo, line 59: Farewell part01, you will be missed 👋

compose/snippets/src/main/java/com/example/compose/snippets/adaptivelayouts/SampleListDetailPaneScaffold.kt
```diff
@@ -105,8 +95,8 @@ fun SampleListDetailPaneScaffoldParts() {
         // [END_EXCLUDE]
         detailPane = {
             AnimatedPane(Modifier) {
```
alexvanyo:
```suggestion
            AnimatedPane {
```

compose/snippets/src/main/java/com/example/compose/snippets/adaptivelayouts/SampleListDetailPaneScaffold.kt
```diff
@@ -49,19 +47,13 @@ import androidx.compose.ui.unit.sp
 @Composable
 fun SampleListDetailPaneScaffoldParts() {
     // [START android_compose_adaptivelayouts_sample_list_detail_pane_scaffold_part02]
-    val navigator = rememberListDetailPaneScaffoldNavigator<Nothing>()
+    val navigator = rememberListDetailPaneScaffoldNavigator<MyItem>()
```
alexvanyo, line 51: Will this correctly save `MyItem` in saved instance state? We aren't using `MyItem.Saver` anywhere now, which seems suspicious.

IanGClifton: You're right, we aren't actually passing the saver into the rememberX so I guess it has to be a Parcelable class to be safely used here.

alexvanyo (approved):
LGTM!

PR #272, opened 2024-05-20
Added snippet for AnimatedVisibility usage with Shared elements

riggaroo:
As well as snippets for CompositionLocal usage. 

https://github.com/android/snippets/assets/9973046/6298fc47-0dd9-4219-87a3-6c9cc431a7ae

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementBlurSnippet.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementBlurSnippet.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 @file:OptIn(ExperimentalSharedTransitionApi::class)
 
 package com.example.compose.snippets.animations.sharedelement
--- a/compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementBlurSnippet.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementBlurSnippet.kt
@@ -1,0 +1,152 @@
+@file:OptIn(ExperimentalSharedTransitionApi::class)
+
+package com.example.compose.snippets.animations.sharedelement
+
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.BoundsTransform
+import androidx.compose.animation.ExperimentalSharedTransitionApi
+import androidx.compose.animation.SharedTransitionLayout
+import androidx.compose.animation.SharedTransitionScope
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.core.tween
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
+import androidx.compose.animation.scaleIn
+import androidx.compose.animation.scaleOut
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.drawWithContent
+import androidx.compose.ui.graphics.BlurEffect
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.TileMode
+import androidx.compose.ui.graphics.layer.GraphicsLayer
+import androidx.compose.ui.graphics.layer.drawLayer
+import androidx.compose.ui.graphics.rememberGraphicsLayer
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.example.compose.snippets.R
+
+private val listSnacks = listOf(
+    Snack("Cupcake", "", R.drawable.cupcake),
+    Snack("Donut", "", R.drawable.donut),
+    Snack("Eclair", "", R.drawable.eclair),
+    Snack("Froyo", "", R.drawable.froyo),
+    Snack("Gingerbread", "", R.drawable.gingerbread),
+    Snack("Honeycomb", "", R.drawable.honeycomb),
+)
+
+private fun <T> animationSpec() = tween<T>(durationMillis = 500)
+private val boundsTransition = BoundsTransform { _, _ -> animationSpec() }
+private val shapeForSharedElement = RoundedCornerShape(16.dp)
+
+@OptIn(ExperimentalSharedTransitionApi::class)
+@Preview
+@Composable
+private fun AnimatedVisibilitySharedElementBlurLayer() {
+    var selectedSnack by remember { mutableStateOf<Snack?>(null) }
+    val graphicsLayer = rememberGraphicsLayer()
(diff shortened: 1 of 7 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/BasicSharedElementSnippets.kt
```diff
@@ -441,3 +443,57 @@ private object UniqueKeySnippet {
...
+            LocalSharedTransitionScope provides this
+        ) {
+            // This could also be your top-level NavHost as this provides an AnimatedContentScope
+            AnimatedContent(state, label = "Top level AnimatedContent") { targetState ->
+                CompositionLocalProvider(LocalNavAnimatedVisibilityScope provides this) {
+                    // Now we can access the scopes in any nested composables as follows:
+                    val sharedTransitionScope = LocalSharedTransitionScope.current
+                        ?: throw IllegalStateException("No SharedElementScope found")
+                    val animatedVisibilityScope = LocalNavAnimatedVisibilityScope.current
+                        ?: throw IllegalStateException("No SharedElementScope found")
```
doris4lt: Should the error mention AnimatedVisibilityScope?

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementSnippets.kt
```diff
@@ -0,0 +1,353 @@
...
+                .fillMaxSize()
+                .background(Color.LightGray.copy(alpha = 0.5f))
+                .padding(16.dp),
+            columns = GridCells.Adaptive(150.dp),
+            horizontalArrangement = Arrangement.spacedBy(8.dp),
+            verticalArrangement = Arrangement.spacedBy(8.dp)
+            // [END_EXCLUDE]
+        ) {
+            itemsIndexed(listSnacks) { index, snack ->
+                AnimatedVisibility(
```
doris4lt, line 94: In a real world use case, I imagine two common use cases for the item after it transitioned to the Details via shared element:

1. Devs may want the item to no longer take space in the list, which means other items would move in. I would expect animateItem to do that once lookahead is properly supported in LazyGrid.

or 2. The item that disappeared in the original list may have a place holder holding its place, so that no movements from other items are needed. This could be achieved by putting AV in a box, with AV's sibling observe the size of _AV's content_ via onSizeChanged, and setting itself to that size. 

For now, since 1 isn't feasible yet, could we try 2 to ensure continuity?

doris4lt: BTW, LazyColumn has better support of lookahead.

We could consider showing an example of AVs in LazyColumn. That way we could set the exit on the AV to shrinkVertically to achieve the use case #1 mentioned above.

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/BasicSharedElementSnippets.kt
riggaroo: yes good catch!

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementSnippets.kt
riggaroo: done - switched to lazy column

compose/snippets/src/main/java/com/example/compose/snippets/landing/LandingScreen.kt
```diff
@@ -16,14 +16,17 @@
...
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
-import androidx.compose.material3.Button
+import androidx.compose.material.Divider
```
MagicalMeghan: Why not Material3 divider here?
https://m3.material.io/components/divider/overview

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementSnippets.kt
```diff
@@ -0,0 +1,358 @@
...
+import androidx.compose.ui.graphics.TileMode
+import androidx.compose.ui.graphics.layer.GraphicsLayer
+import androidx.compose.ui.graphics.layer.drawLayer
+import androidx.compose.ui.graphics.rememberGraphicsLayer
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.example.compose.snippets.R
+
```
MagicalMeghan, line 64: Is this file only for one snippet? It looks like there are a lot of parts needed to have this code work - is there enough context in the documentation for the developer to understand everything needed here?

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/BasicSharedElementSnippets.kt
```diff
@@ -441,3 +443,57 @@ private object UniqueKeySnippet {
...
 }
+
+// [START android_compose_shared_element_scope]
+val LocalNavAnimatedVisibilityScope = compositionLocalOf<AnimatedVisibilityScope?> { null }
+val LocalSharedTransitionScope = compositionLocalOf<SharedTransitionScope?> { null }
+
+@Composable
+private fun SharedElementScope_CompositionLocal() {
+    // An example of how to use composition locals to pass around the shared transition scope, far down your UI tree.
+    // [START_EXCLUDE]
```
MagicalMeghan, line 454: Since youre giving the entire method here does it make sense to exclude these parts? If a developer copy/pastes this in they may be confused about where state should be defined

compose/snippets/src/main/java/com/example/compose/snippets/landing/LandingScreen.kt
```diff
@@ -82,9 +84,14 @@ fun NavigationItems(navigate: (Destination) -> Unit) {
 
 @Composable
 fun NavigationItem(destination: Destination, onClick: () -> Unit) {
-    Button(
-        onClick = { onClick() }
+    Box(
```
MagicalMeghan, line 87: Why a box instead of a button here?

riggaroo: It looks better than a list of buttons :)

compose/snippets/src/main/java/com/example/compose/snippets/landing/LandingScreen.kt
riggaroo: good catch!

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/BasicSharedElementSnippets.kt
riggaroo: The snippet is in context of teaching about CompositionLocal usage with SharedTransitionLayout + AnimatedContent, the expectataion is the the developer would know what state they want to provide to their AnimatedContent composable.

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementSnippets.kt
riggaroo: Fair point, it had an example with the blurring and without blurring, i've moved the blurring full snippet to another snippet file that wont be referenced on DAC. and simplified this one.

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementBlurSnippet.kt
```diff
@@ -0,0 +1,168 @@
...
+    Snack("Eclair", "", R.drawable.eclair),
+    Snack("Froyo", "", R.drawable.froyo),
+    Snack("Gingerbread", "", R.drawable.gingerbread),
+    Snack("Honeycomb", "", R.drawable.honeycomb),
+)
+
+private fun <T> animationSpec() = tween<T>(durationMillis = 500)
+private val boundsTransition = BoundsTransform { _, _ -> animationSpec() }
+private val shapeForSharedElement = RoundedCornerShape(16.dp)
+
```
MagicalMeghan, line 70: Why have this in snippets if its not going on DAC?

MagicalMeghan (approved):
LGTM - just one question

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementBlurSnippet.kt
riggaroo: To be able to point people to it as an advanced example.

riggaroo: ie from Twitter or a video etc.

PR #273, opened 2024-05-20
Add an example of seeking between shared element states manually using SeekableTransitionState

riggaroo:
Using `SeekableTransitionState` and `rememberUpdatedTransition()` with `Transition#AnimatedContent`, we can control the shared element seeking state,

In this example, a Slider is added to allow for seeking between Home and Details state. The slider also shows progress as the animation happens - so when you click on an item, the progress bar will animate from 0 to 1 - and then back to 0 as it completes the transition. 

`SeekableTransitionState` can be used to build up your own navigation library, or between different states within one Screen of your UI, if you need the ability to seek between states.


compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/SharedElementsWithNavigationSnippets.kt
```diff
@@ -193,3 +215,85 @@ data class Snack(
...
+    val seekableTransitionState = remember {
+        SeekableTransitionState<Screen>(Screen.Home)
+    }
+    val transition = rememberTransition(transitionState = seekableTransitionState)
+
+    PredictiveBackHandler(seekableTransitionState.currentState is Screen.Details) { progress ->
+        try {
+            progress.collect { backEvent ->
+                // code for progress
+                seekableTransitionState.seekTo(backEvent.progress, targetState = Screen.Home)
```
doris4lt, line 238: seekTo may be cancelled _as expected_, if `animateTo` or subsequent `seekTo` gets called before the current `seekTo` finishes. See the mutatorMutex shared between these calls: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/animation/animation-core/src/commonMain/kotlin/androidx/compose/animation/core/Transition.kt;l=507

You may need a try-catch around this seekTo as well to ignore the cancellation from seekTo.

jbw0033 (approved):

PR #276, opened 2024-05-24
Adding basic badge examples

jakeroseman:


compose/snippets/src/main/java/com/example/compose/snippets/components/Badges.kt
```diff
@@ -0,0 +1,74 @@
...
+            imageVector = Icons.Filled.ShoppingCart,
+            contentDescription = "Shopping cart"
+        )
+    }
+}
+
+@Composable
+fun BadgeNumberExample() {
+    BadgedBox(
+        badge = { Text("8") }
```
arriolac: Should the Text be wrapped within a `Badge` composable?

jakeroseman: Yes, just fleshing this out now. Forgot to mark this as a draft!

PR #277, opened 2024-05-24
Adding badge examples

jakeroseman:
Adding examples for the badge component page

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/components/Badges.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/components/Badges.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/components/Badges.kt
```diff
@@ -0,0 +1,115 @@
...
+fun BadgeInteractiveExample() {
+    var itemCount by remember { mutableStateOf(0) }
+
+    Column(
+        verticalArrangement = Arrangement.spacedBy(16.dp)
+    ) {
+        if (itemCount > 0) {
+            BadgedBox(
+                badge = {
+                    Badge(
```
arriolac, line 89: Should the if condition be here instead? So you would still have a BadgedBox container even if itemCount == 0

jakeroseman: Done!

PR #280, opened 2024-06-10
Adds Wear module and voice input snippet

kul3r4:
- Adds a Wear module
- Adds a snippet to replace outdated code in https://developer.android.com/training/wearables/user-input/voice#kotlin


wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
```diff
@@ -0,0 +1,103 @@
...
+    var textForVoiceInput by remember { mutableStateOf("") }
+
+    val voiceLauncher =
+        rememberLauncherForActivityResult(
+            ActivityResultContracts.StartActivityForResult()
+        ) {
+            // This is where you process the intent and extract the speech text from the intent.
+            it.data?.let { data ->
+                val results = data.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)
+                textForVoiceInput = results?.get(0) ?: "None"
```
garanj, line 76: Would this be better as a nullable string, instead of using "None"?

wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+    val scrollState = rememberScrollState()
+
+    ScreenScaffold(scrollState = scrollState) {
```
garanj, line 83: For the purposes of this snippet, is it possible to create something that doesn't involve including the Scaffold and Column code but still illustrates the point? In the same way that the existing snippet for voice doesn't include UI code? (Just thinking that this gets a bit long).

kul3r4: good point, added a couple of `START_EXCLUDE android_wear_voice_input]` and `[END_EXCLUDE android_wear_voice_input]` to remove some code

wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.google.android.horologist.compose.layout.ScreenScaffold
+import com.google.android.horologist.compose.material.Chip
+import com.google.android.horologist.compose.rotaryinput.rotaryWithScroll
+
+/**
+ * Shows voice input option
+ */
+@OptIn(ExperimentalHorologistApi::class)
+@Composable
+fun VoiceInputScreen(
```
riggaroo: Ensure you are running spotless on this module and that its also added to github build actions to ensure spotless has run and it compiles correctly.

wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
```diff
@@ -0,0 +1,108 @@
...
+) {
+    // [START android_wear_voice_input]
+    var textForVoiceInput by remember { mutableStateOf("") }
+
+    val voiceLauncher =
+        rememberLauncherForActivityResult(
+            ActivityResultContracts.StartActivityForResult()
+        ) {
+            // This is where you process the intent and extract the speech text from the intent.
+            it.data?.let { data ->
```
riggaroo: `it` in this context for me doesn't have much meaning. Is it possible to rename to what its representing?

wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
riggaroo: I dont think you need the android_wear_voice_input tag, you can just use start and end_exclude

wear/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<resources>
+    <string name="app_name">My Application</string>
```
riggaroo: Might want to give this a better name

wear/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,7 @@
+<resources>
+
+    <style name="MainActivityTheme.Starting" parent="Theme.SplashScreen">
```
riggaroo: Can you name this WearActivityTheme? Or WearTheme? Also why is it "Starting"?

riggaroo (changes requested):
Needs the build scripts and spotless :) Otherwise its looking great!

wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
kul3r4: Done thanks! https://github.com/android/snippets/pull/280/files#diff-7e266b7fa3927fbdcff448f786d1c8c9bd7b918620cfff9adf59d2c7aeabff9fR50

wear/src/main/java/com/example/wear/snippets/voiceinput/VoiceInputScreen.kt
kul3r4: renamed to `activityResult`

wear/src/main/res/values/strings.xml
kul3r4: Renamed to "Wear Snippets"

wear/src/main/res/values/styles.xml
kul3r4: Good catch, we are actually fine with default theme on Wear since Splashscreen is taken care by the platform. So this file was removed.

riggaroo (approved):

ABirHAsan123-beep (approved):
3ae1a77fc85c48a8ca423a58708e5e8117ee2c83

PR #284, opened 2024-06-21
Adds sample for rotary and timepicker

kul3r4:
This PR partially fixes https://developer.android.com/training/wearables/compose/rotary-input:
- Adds an example for a scrollable screen where rotary can be used to scroll
- Adds an example for a TimePicker to demonstrate how to use rotary on a Picker + focus

A follow up PR will add the remaining part of the DAC page.

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/rotary/Rotary.kt
+++ b/wear/src/main/java/com/example/wear/snippets/rotary/Rotary.kt
@@ -1,0 +1,219 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


wear/src/main/java/com/example/wear/snippets/rotary/Rotary.kt
```diff
@@ -0,0 +1,219 @@
...
+
+@OptIn(ExperimentalComposeUiApi::class)
+@Composable
+fun TimePicker() {
+    val textStyle = MaterialTheme.typography.display1
+
+    // [START android_wear_rotary_input_picker]
+    var selectedColumn by remember { mutableIntStateOf(0) }
+
+    val focusRequester1 = remember { FocusRequester() }
```
garanj: Are focusRequester1 and 2 just one per `Picker` component? Wondering if there is a better name...?

garanj (approved):

wear/src/main/java/com/example/wear/snippets/rotary/Rotary.kt
kul3r4: good point, renamed to  `minutesFocusRequester` and `hoursFocusRequester`

PR #286, opened 2024-07-03
Adds code for List using Horologist

kul3r4:
Adds sample code for [Wear Compose list DAC](https://developer.android.com/training/wearables/compose/lists) using the `ScalingLazyColumn` from Horologist. The page itself will need a rewrite after adding this snippet.

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/list/List.kt
+++ b/wear/src/main/java/com/example/wear/snippets/list/List.kt
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


wear/src/main/java/com/example/wear/snippets/list/List.kt
```diff
@@ -0,0 +1,110 @@
...
+                )
+            }
+        }
+    }
+    // [END android_wear_list]
+}
+
+@OptIn(ExperimentalHorologistApi::class)
+@Composable
+fun SnapAndFlingComposeList() {
```
garanj, line 68: Which is the Fling part?

garanj (approved):
Looks good! I think maybe I don't understand snap and fling very well, I can see the rotaryMode set to Snap, where does the fling come into things?

wear/src/main/java/com/example/wear/snippets/list/List.kt
kul3r4: Actually you spotted a problem: fling has been renamed to Scroll, so the correct way is to use `rotaryMode = ScalingLazyColumnState.RotaryMode.Scroll` which I just updated.

kul3r4: Actually, `rotaryMode = ScalingLazyColumnState.RotaryMode.Scroll` is the default already.

kul3r4: By looking at https://github.com/google/horologist/blob/main/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt#L198, if you specify Snap you get the Snap and fling

PR #300, opened 2024-07-25
Latest beta06 updates for Compose

riggaroo:
Upgrade kotlin 2.0.0 and apply compose compiler plugin


bentrengrove (approved):
All fine except that weird test that I don't understand how it didn't fail previously?

PR #320, opened 2024-08-27
Migrate many snippets to use Material 3 instead of Material 2. 

riggaroo:
Aligning more of snippets with the recommendations to use Material 3 instead of Material 2.


compose/snippets/src/main/java/com/example/compose/snippets/graphics/ShapesSnippets.kt
```diff
@@ -372,12 +373,14 @@ class RoundedPolygonShape(
...
-            rounding = CornerRounding(0.2f)
+        mutableStateOf(
+            RoundedPolygon(
+                6,
+                rounding = CornerRounding(0.2f)
+            )
         )
     }
-    val clip = remember(hexagon) {
+    val clip = remember {
```
bentrengrove, line 383: I don't think your clip will update with a new shape without the key on the remember

compose/snippets/src/main/java/com/example/compose/snippets/text/TextDownloadableFontsSnippets.kt
```diff
@@ -200,6 +200,11 @@ private object DownloadableFontsDebugSnippet {
     // [END android_compose_text_df_debug_apis_handler]
 }
 
+/**
```
bentrengrove: What's this for?

bentrengrove (approved):

compose/snippets/src/main/java/com/example/compose/snippets/text/TextDownloadableFontsSnippets.kt
riggaroo: oops, accidental commit

PR #340, opened 2024-09-11
Docs: Adds a Glide snippet for the "Loading Images" Quick Guide at

thedmail:
https://developer.android.com/quick-guides/content/load-images?hl=en

### DIFF
--- a/compose/snippets/build.gradle.kts
+++ b/compose/snippets/build.gradle.kts
@@ -159,5 +159,6 @@ dependencies {
 
     androidTestImplementation(libs.androidx.compose.ui.test.junit4)
     debugImplementation(libs.androidx.compose.ui.test.manifest)
+    implementation("com.github.bumptech.glide:compose:1.0.0-beta01")
 
 }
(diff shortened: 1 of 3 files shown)


compose/snippets/build.gradle.kts
```diff
@@ -159,5 +159,6 @@ dependencies {
 
     androidTestImplementation(libs.androidx.compose.ui.test.junit4)
     debugImplementation(libs.androidx.compose.ui.test.manifest)
+    implementation("com.github.bumptech.glide:compose:1.0.0-beta01")
```
riggaroo: Please add the dependency to the libs.toml file and then reference the dependency like above libraries instead.

thedmail: Thanks. Will do.

thedmail: Tried to do it by imitating the way it's done for coil. Didn't work.

Is there documentation for what you're asking for?

thedmail: Think I got it figured out. It builds on my machine now. :-)

riggaroo (approved):

PR #347, opened 2024-09-16
Validate input

thedmail:


compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
```diff
@@ -839,3 +840,55 @@ private val firaSansFamily = FontFamily()
...
+        }
+    }
+
+    @Composable
+    fun ValidatingInputTextField(
+        email: String,
+        updateState: (String) -> Unit,
+        validatorHasErrors: Boolean
+    ) {
+        val emailViewModel = EmailViewModel()
```
riggaroo: This should use the recommended pattern here: https://developer.android.com/develop/ui/compose/libraries#viewmodel

i.e. import ` implementation 'androidx.lifecycle:lifecycle-viewmodel-compose:2.8.5'`
and then `viewModel: MyViewModel = viewModel()`

riggaroo (changes requested):
Update viewmodel usage

compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
thedmail: Doesn't work when I do that. If I try it in the function signature, I get an "expecting an element" error. (There's nothing to pass, which is why we're declaring the variable locally.)

But if I try declaring it outside the signature, as a local variable, based on the pattern in the doc:

val emailViewModel: EmailViewModel=emailViewModel()

I get an error that emailViewModel is undeclared.

riggaroo: fixed it for you

riggaroo (approved):

compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
thedmail: ty!

PR #350, opened 2024-09-17
Lazily load list

thedmail:


compose/snippets/src/main/java/com/example/compose/snippets/lists/LazyListSnippets.kt
```diff
@@ -742,3 +750,53 @@ private val randomSizedPhotos = listOf(
...
+        modifier: Modifier,
+        message: Text
+    ) {
+        Card(modifier = Modifier.padding(8.dp)) {
+            Column(
+                modifier = Modifier.padding(8.dp),
+                verticalArrangement = Arrangement.Center
+            ) {
+                // Text(message.sender), where "message" is an object with a sender property
+                // Text(message.text), where "text" is an object with a text property
```
riggaroo: Should be uncommented?

riggaroo (changes requested):
Uncomment the Text composables.

compose/snippets/src/main/java/com/example/compose/snippets/lists/LazyListSnippets.kt
thedmail: No. Because there's no actual message class anywhere that has a sender or text property, and this example was too complex for me to set something up idiomatically (at least in a time-efficient way).

If you feel it's necessary to have this in code, could I ask you to implement it?

riggaroo: Creating a simple class to represent the message structure is a common practice in DAC snippets. It helps make the code compile and snippets readable instead of the code being commented out. 

While it's not strictly necessary for this particular example as you could just put a `Text("Lazy list text")`, it could be a good opportunity to practice this pattern, as it's frequently used in other snippets and real-world scenarios.

Would you be open to giving it a try? I'm happy to provide guidance and support if you need it.

Something like a dummy class like this would match this fake class:

```data class Message(sender: String, text: String)```

thedmail: Sure--thank you for helping me learn!!

There's actually a reason why I didn't go ahead and create a data class as you're suggesting. What has me a little mixed-up are these lines:

lazyPagingItems = pager.flow.collectAsLazyPagingItems()

message = lazyPagingItems[index]

So I don't see how I can make message be an object of a new data class (such as I'd create) when it's already a variable set equal  of  LazyPagingItems type.

This was the complexity that stopped me from just defining my own class, and instead  had me resort to commenting.

riggaroo: fixed it for you

riggaroo (approved):

compose/snippets/src/main/java/com/example/compose/snippets/lists/LazyListSnippets.kt
thedmail: tyvm

PR #357, opened 2024-09-27
Password update to use BasicSecureTextField

riggaroo:
https://github.com/user-attachments/assets/0739da86-22d4-4670-ad09-9bdc9980fd07


compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
```diff
@@ -862,31 +867,43 @@ val Purple = Color(0xFF800080)
...
-                    Icons.Filled.VisibilityOff
-                },
-                contentDescription = "Toggle password visibility",
-                modifier = Modifier.clickable { showPassword = !showPassword }
-            )
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(6.dp)
+            .border(1.dp, Color.LightGray, RoundedCornerShape(6.dp))
+            .padding(6.dp),
```
dturner, line 884: Looks like the padding modifier is set twice.

dturner (approved):

compose/snippets/src/main/java/com/example/compose/snippets/text/TextSnippets.kt
riggaroo: This is intentional to add the padding after and before the border :)

PR #368, opened 2024-10-04
Snippets for keyboard input

chikoski:
This pull request adds following snippets:

- Snippets for [Keyboard Shortcuts Helper](https://developer.android.com/develop/ui/compose/touch-input/keyboard-input/keyboard-shortcuts-helper)
- Snippets for "Handle Keyboard Actions"

The second one is for a new doc.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/touchinput/keyboardinput/KeyboardShortcutsHelper.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/touchinput/keyboardinput/KeyboardShortcutsHelper.kt
@@ -66,9 +66,13 @@ class AnotherActivity : ComponentActivity() {
         setContent {
             MaterialTheme {
                 // [START android_compose_keyboard_shortcuts_helper_request]
-                val activity = LocalContext.current as Activity
+                val activity = LocalContext.current as? Activity
 
-                Button(onClick = { activity.requestShowKeyboardShortcuts() }) {
+                Button(
+                    onClick = {
+                        activity?.requestShowKeyboardShortcuts()
+                    }
+                ) {
                     Text(text = "Show keyboard shortcuts")
                 }
                 // [END android_compose_keyboard_shortcuts_helper_request]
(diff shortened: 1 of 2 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/touchinput/keyboardinput/KeyboardShortcutsHelper.kt
```diff
@@ -0,0 +1,108 @@
...
+class AnotherActivity : ComponentActivity() {
+
+    @RequiresApi(Build.VERSION_CODES.N)
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        enableEdgeToEdge()
+        setContent {
+            MaterialTheme {
+                // [START android_compose_keyboard_shortcuts_helper_request]
+                val activity = LocalContext.current as Activity
```
bentrengrove: You can't cast context to activity like this because there are cases where it won't be one. Let me see what I can find out to help you do this better

chikoski: Thanks for catching that. To explain the API, the following code is sufficient. What do you think? 

```Kotlin

Button(
    onClick = {
        if(context is Activity){
             context.requestShowKeyboardShortcuts()
        }
    }
) {
    Text("Show shortcuts")
}
```

bentrengrove: I think just `LocalContext.current as? Activity`

and then

`activity?.requestShowKeyboardShortcuts()`

is cleaner, but no strong opinion if you think otherwise

chikoski: Thanks for the suggestion. Updated the code as suggested as the suggested one seems simpler :)

PR #373, opened 2024-10-11
Tooltip component examples

jakeroseman:
Two basic example and one advanced example of how to implement the tooltip material component

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
@@ -94,7 +94,7 @@ fun PlainTooltipExample(
         IconButton(onClick = { /* Do something... */ }) {
             Icon(
                 imageVector = Icons.Filled.Favorite,
-                contentDescription = "Localized Description"
+                contentDescription = "Add to favorites"
             )
         }
     }
@@ -194,7 +194,7 @@ fun AdvancedRichTooltipExample(
         IconButton(onClick = { tooltipState.dismiss() }) {
             Icon(
                 imageVector = Icons.Filled.Camera,
-                contentDescription = "Localized Description"
+                contentDescription = "Open camera"
             )
         }
     }
--- a/compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
@@ -1,0 +1,162 @@
+package com.example.compose.snippets.components
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Camera
+import androidx.compose.material.icons.filled.Favorite
+import androidx.compose.material.icons.filled.Info
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
+import androidx.compose.material3.PlainTooltip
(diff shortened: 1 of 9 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
```diff
@@ -0,0 +1,177 @@
...
+        positionProvider = TooltipDefaults.rememberPlainTooltipPositionProvider(),
+        tooltip = {
+            PlainTooltip { Text(plainTooltipText) }
+        },
+        state = rememberTooltipState()
+    ) {
+        IconButton(onClick = { /* Icon button's click event */ }) {
+            Icon(
+                imageVector = Icons.Filled.Favorite,
+                contentDescription = "Localized Description"
```
JolandaVerhoef: What about making this `plainTooltipText` as well? "Localized Description" does not really work for blind people to understand what's going on :)

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
```diff
@@ -0,0 +1,177 @@
...
+                title = { Text(richTooltipSubheadText) }
+            ) {
+                Text(richTooltipText)
+            }
+        }, state = tooltipState
+    ) {
+        IconButton(onClick = { /* Icon button's click event */ }) {
+            Icon(
+                imageVector = Icons.Filled.Info,
+                contentDescription = "Localized Description"
```
JolandaVerhoef: same here

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
```diff
@@ -0,0 +1,177 @@
...
+        tooltip = {
+            RichTooltip(
+                title = { Text(richTooltipSubheadText) },
+                action = {
+                    Row {
+                        TextButton(onClick = { /* Do something... */ }) {
+                            Text(richTooltipActionText)
+                        }
+                        TextButton(onClick = { /* Do something... */ }) {
+                            Text("Learn more")
```
JolandaVerhoef: Why is one hard-coded, and the other passed as a parameter?

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
```diff
@@ -0,0 +1,177 @@
...
+
+    TooltipBox(
+        modifier = modifier,
+        positionProvider = TooltipDefaults.rememberRichTooltipPositionProvider(),
+        tooltip = {
+            RichTooltip(
+                title = { Text(richTooltipSubheadText) },
+                action = {
+                    Row {
+                        TextButton(onClick = { /* Do something... */ }) {
```
JolandaVerhoef: I think we can make the dismiss action work, that would be nice

JolandaVerhoef (changes requested):
There's a problem with the behavior of the rich tooltips. After I open them and then dismiss them, I can't open them again. Except when I first show another tooltip.

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
jakeroseman: Well observed. I've just removed the second one.

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
jakeroseman: Done.

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
jakeroseman: Done.

compose/snippets/src/main/java/com/example/compose/snippets/components/Tooltips.kt
jakeroseman: Done.

JolandaVerhoef (approved):

PR #380, opened 2024-10-14
Navigation drawer examples

jakeroseman:


compose/snippets/src/main/java/com/example/compose/snippets/components/NavigationDrawer.kt
```diff
@@ -0,0 +1,195 @@
...
+                Text("Navigation drawer with nested items.")
+            }
+        }
+
+        return
+    }
+
+    Column(modifier = Modifier.fillMaxSize()){
+        Text(
+            text = "Swipe from left to open the ${currentExample!!.description}.",
```
JolandaVerhoef: Instead of using `!!` you could instead nest this in the else part of the `if(currentExample == null)` statement.

compose/snippets/src/main/java/com/example/compose/snippets/components/NavigationDrawer.kt
```diff
@@ -0,0 +1,195 @@
...
+
+@Preview
+@Composable
+private fun NavigationDrawerExamplesPreview(){
+    NavigationDrawerExamples()
+}
+
+// [START android_compose_components_navigationdrawergroupitem]
+@Composable
+fun NavigationDrawerGroupItem(
```
JolandaVerhoef: Move this down to closer to the nested example

compose/snippets/src/main/java/com/example/compose/snippets/components/NavigationDrawer.kt
jakeroseman: The linter still insists on the `!!` even when in an `else` block. I don't think it's a problem here as we return early on null.

compose/snippets/src/main/java/com/example/compose/snippets/components/NavigationDrawer.kt
jakeroseman: Done.

JolandaVerhoef (approved):

PR #382, opened 2024-10-16
Add basic Camera sample

JolandaVerhoef:


media/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+
+dependencies {
+
+    // Add CameraX Compose dependency
+    implementation(libs.androidx.camera.compose)
+    implementation("androidx.camera:camera-core:1.3.4")
+    implementation("androidx.camera:camera-camera2:1.3.4")
+
+    implementation(libs.androidx.core.ktx)
+    implementation(libs.androidx.lifecycle.runtime)
```
temcguir, line 51: You'll also need `implementation(libs.androidx.lifecycle.viewModelCompose)`

media/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
```
temcguir, line 4: Need camera feature / permission:
```
    <uses-feature
        android:name="android.hardware.camera"
        android:required="true" />
    <uses-feature
        android:name="android.hardware.camera.autofocus"
        android:required="false" />

    <uses-permission android:name="android.permission.CAMERA" />
```

temcguir: You'll also need some way of granting permissions before launching the camera.

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,89 @@
...
+        viewModelScope.launch {
+            val imageProxy = captureUseCase.takePicture()
+            // Do something with the image
+        }
+    }
+}
+
+@Composable
+fun CameraPreviewScreen(modifier: Modifier = Modifier) {
+    val viewModel = CameraPreviewViewModel(LocalContext.current.applicationContext)
```
temcguir: ViewModel should be retrieved via `androidx.lifecycle.viewmodel.compose.viewModel`. Something like:
```
@Composable
fun CameraPreviewScreen(
    modifier: Modifier = Modifier,
    viewModel: CameraPreviewViewModel = viewModel()
) {
    val surfaceRequest by viewModel.surfaceRequest.collectAsStateWithLifecycle()
```

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,89 @@
...
+    val surfaceRequest by viewModel.surfaceRequest.collectAsStateWithLifecycle()
+
+    LifecycleStartEffect(Unit) {
+        viewModel.startCamera()
```
temcguir: We can just pass the context to the camera here instead of through the ViewModel constructor. This is to align with the advice from https://developer.android.com/topic/architecture/recommendations#viewmodel where it says not to use `AndroidViewModel`. Though I wonder if there is still a better way to pass the context in.

```
    LifecycleStartEffect(Unit) {
        viewModel.startCamera(context)
        onStopOrDispose {
            viewModel.stopCamera()
        }
    }
```

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,89 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+
+class CameraPreviewViewModel(private val appContext: Context): ViewModel() {
+    private val _surfaceRequest = MutableStateFlow<SurfaceRequest?>(null)
+    val surfaceRequest: StateFlow<SurfaceRequest?> = _surfaceRequest
+
+    private lateinit var processCameraProvider: ProcessCameraProvider
+    private val previewUseCase = Preview.Builder().build()
```
temcguir: You should probably just add the surfaceProvider here:
```
    private val previewUseCase = Preview.Builder().build().apply {
        setSurfaceProvider { newSurfaceRequest ->
            _surfaceRequest.value = newSurfaceRequest
        }
    }
```

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,89 @@
...
+    private var runningCameraJob: Job? = null
+
+    fun startCamera() {
+        stopCamera()
+        runningCameraJob = viewModelScope.launch {
+            processCameraProvider = ProcessCameraProvider.awaitInstance(appContext)
+            processCameraProvider.runWith(CameraSelector.DEFAULT_BACK_CAMERA, useCaseGroup) {
+                previewUseCase.setSurfaceProvider { newSurfaceRequest ->
+                    _surfaceRequest.value = newSurfaceRequest
+                }
```
temcguir: This is normally where we would suspend and wait for camera setting updates, such as zoom or tap-to-focus events. Since we're not doing any of that, you'll need an `awaitCancellation()` here.

temcguir (changes requested):
Some changes to make in order to get it working with the current `CoroutineLifecycleOwner` code. I'll take another pass to get it working without `CoroutineLifecycleOwner`.

media/build.gradle.kts
JolandaVerhoef: I don't think the very basic implementation would actually need a "real" view model.

temcguir: If we're going to have a `ViewModel` class, why wouldn't it be real? The only difference between the current code  and a real ViewModel is the lifecycle of the ViewModel. Currently, the code is recreating the ViewModel on every composition. That means all the state from the previous composition will be lost. That won't work for a variety of reasons. [This comment](https://github.com/android/snippets/pull/382#discussion_r1807063767) provides a single-line change that would turn it into a "real" ViewModel.

I suppose another option would be to just put the logic into a plain `StateHolder` object, but it would be roughly the same amount of code, and since the `ViewModel` in this sample contains "business logic" (`startCamera`, `stopCamera`, `takePicture`) , it seems like it lines up with [this documentation](https://developer.android.com/develop/ui/compose/state-hoisting#business-logic) that recommends using a ViewModel.

JolandaVerhoef: If we'd want to show a real-world copyable example, we'd have to use Hilt and inject a ProcessCameraProvider into the view model (which itself depends on app context). 

Showing a half-baked solution where we pass the context in the start/stop camera is not what I'd recommend. Instead, I'd leave the view model instantiation out of the snippets all-together. I updated the PR to reflect that, moving the ViewModel creation into the activity, which wouldn't be part of the snippets. 

Then, in the documentation, we can simply mention that the view model should be injected using dependency injection.

Happy to discuss this further in our sync :)

temcguir: Sounds fine to me.

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    fun stopCamera() {
+        runningCameraJob?.apply {
+            if (isActive) {
+                cancel()
+            }
+        }
+    }
```
temcguir: With `bindToLifecycle` this will look something like this:
```
    fun startCamera(lifecycleOwner: LifecycleOwner) {
        val prevCameraJob = runningCameraJob
        runningCameraJob = viewModelScope.launch {
            prevCameraJob?.cancelAndJoin()
            val processCameraProvider = ProcessCameraProvider.awaitInstance(appContext)
            processCameraProvider.bindToLifecycle(
                lifecycleOwner,
                CameraSelector.DEFAULT_BACK_CAMERA,
                useCaseGroup
            )

            try {
                awaitCancellation()
            } finally {
                processCameraProvider.unbindAll()
            }
        }
    }

    fun stopCamera() {
        runningCameraJob?.cancel()
    }
```

I've also made `processCameraProvider` a local variable since it's only used here. Since it's cached with the process, there isn't a need to save it as a instance variable.

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,128 @@
...
+// [START android_media_camera_preview_content]
+@Composable
+fun CameraPreviewContent(
+    viewModel: CameraPreviewViewModel,
+    modifier: Modifier = Modifier
+) {
+    val surfaceRequest by viewModel.surfaceRequest.collectAsStateWithLifecycle()
+
+    LifecycleStartEffect(Unit) {
+        viewModel.startCamera()
```
temcguir: With the `lifecycleOwner` added to `startCamera()`, this will be changed to:
```
viewModel.startCamera(this@LifecycleStartEffect)
```

media/src/main/java/com/example/media/camera/CameraSnippets.kt
```diff
@@ -0,0 +1,160 @@
...
+        }.also { runningCameraJob = it }
+    }
+
+    fun stopCamera() {
+        runningCameraJob?.cancel()
+    }
+
+    fun takePicture() {
+        viewModelScope.launch {
+            val imageProxy = captureUseCase.takePicture()
```
MagicalMeghan, line 90: Take picture doesn't compile without parameters passed in

JolandaVerhoef: Can't reproduce on my side. Code works well both with camerax-compose alpha02 and alpha06

temcguir: This is an extension function, so make sure you `import androidx.camera.core.takePicture`.

PR #401, opened 2024-11-27
Add single option radio button example

jakeroseman:
Add basic radio button example to serve the CUJ "Show a single select option to my users."


.kotlin/sessions/kotlin-compiler-5020797412157966223.salive
JolandaVerhoef, line 1: you probably don't want to check this file in

JolandaVerhoef (approved):

.kotlin/sessions/kotlin-compiler-5020797412157966223.salive
jakeroseman: Definitely not. Not sure what's up here.

PR #416, opened 2024-12-10
Add an example of Shared elements that are seekable using SeekableTransitionState

riggaroo:
Added an example of seekable transitions, was previously approved by Doris but pull request got outdated. #273

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/CustomizeSharedElementsSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/CustomizeSharedElementsSnippets.kt
@@ -14,10 +14,13 @@
  * limitations under the License.
  */
 
-@file:OptIn(ExperimentalSharedTransitionApi::class)
+@file:OptIn(ExperimentalSharedTransitionApi::class, ExperimentalSharedTransitionApi::class,
+    ExperimentalSharedTransitionApi::class
+)
 
 package com.example.compose.snippets.animations.sharedelement
 
+import androidx.activity.compose.PredictiveBackHandler
 import androidx.compose.animation.AnimatedContent
 import androidx.compose.animation.AnimatedVisibility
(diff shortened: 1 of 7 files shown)


mlykotom (commented):
should we add `android:enableOnBackInvokedCallback="true"` to the android manifest? otherwise the predictive back doesn't work (I don't know if we don't break some other snippet though)

mlykotom:
When I cancel the back gesture, then it doesn't work again :(

https://github.com/user-attachments/assets/86cd86f1-b4b7-4604-83d5-1f6a3faa1896

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/CustomizeSharedElementsSnippets.kt
```diff
@@ -628,3 +636,91 @@ fun PlaceholderSizeAnimated_Demo() {
...
+                } catch (e: CancellationException) {
+                    // ignore the cancellation
+                }
+            }
+            // code for completion
+            seekableTransitionState.animateTo(seekableTransitionState.targetState)
+        } catch (e: CancellationException) {
+            // code for cancellation
+            seekableTransitionState.animateTo(seekableTransitionState.currentState)
+        }
```
mlykotom: I find it slightly confusing why do I need to suppress the cancellation exceptions here.
Can we explain it a bit more in the comments? 🙏🏼

riggaroo: Added comments.

riggaroo:
> When I cancel the back gesture, then it doesn't work again :(
> 
>  broken.mp4

Fixed this by changing to snapTo instead of animateTo in the cancellation block, could you try it again?

mlykotom (approved):
Thank you, so much clearer with the comments. I tried it on the phone and it's working ❤️

PR #419, opened 2024-12-11
Replaced deprecated WindowHeightSizeClass with isHeightAtLeastBreakpoint().

JonEckenrode:
Updated snippet for window size class documentation.


compose/snippets/src/main/java/com/example/compose/snippets/layouts/AdaptiveLayoutSnippets.kt
```diff
@@ -52,9 +51,9 @@ fun MyApp(
     windowSizeClass: WindowSizeClass = currentWindowAdaptiveInfo().windowSizeClass
 ) {
     // Perform logic on the size class to decide whether to show the top app bar.
-    val showTopAppBar = windowSizeClass.windowHeightSizeClass != WindowHeightSizeClass.COMPACT
+    val showTopAppBar = windowSizeClass.isHeightAtLeastBreakpoint(480)
```
alexvanyo: Let's use the built-in constant in `androidx.window`: https://developer.android.com/reference/androidx/window/core/layout/WindowSizeClass?hl=en#HEIGHT_DP_MEDIUM_LOWER_BOUND()
```suggestion
    val showTopAppBar = windowSizeClass.isHeightAtLeastBreakpoint(WindowSizeClass.HEIGHT_DP_MEDIUM_LOWER_BOUND)
```

Harry19941994 (approved):

compose/snippets/src/main/java/com/example/compose/snippets/layouts/AdaptiveLayoutSnippets.kt
JonEckenrode: Revised.

PR #444, opened 2025-01-28
Adding Autofill snippets

MagicalMeghan:
Adding Autofill snippets for documentation

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
@@ -1,0 +1,122 @@
+package com.example.compose.snippets.text
+
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.text.BasicTextField
+import androidx.compose.foundation.text.LocalAutofillHighlightColor
+import androidx.compose.foundation.text.input.TextFieldState
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.autofill.ContentType
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalAutofillManager
+import androidx.compose.ui.semantics.contentType
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.unit.dp
+import com.example.compose.snippets.touchinput.Button
+
+@Composable
+fun AddAutofill() {
+    // [START android_compose_autofill_1]
+    BasicTextField(
+        state = remember { TextFieldState("Enter your username.") },
+        modifier = Modifier.semantics { contentType = ContentType.Username }
+    )
+    // [END android_compose_autofill_1]
+}
+
+@Composable
+fun AddMultipleTypesOfAutofill() {
+    // [START android_compose_autofill_2]
+    Column {
+        BasicTextField(
+            state = remember { TextFieldState() },
+            modifier =
+            Modifier.semantics {
+                contentType = ContentType.Username + ContentType.EmailAddress
+            },
+        )
+    }
+    // [END android_compose_autofill_2]
+}
+
+@Composable
+fun SaveDataWithAutofill() {
+    // [START android_compose_autofill_3]
+    // [START android_compose_autofill_4]
+    val autofillManager = LocalAutofillManager.current
+    // [END android_compose_autofill_3]
+
+    Column {
+        BasicTextField(
+            state = remember { TextFieldState() },
+            modifier =
+            Modifier.semantics {
+                contentType = ContentType.NewUsername
+            },
+        )
+
+        Spacer(modifier = Modifier.height(16.dp))
+
+        BasicTextField(
+            state = remember { TextFieldState() },
+            modifier =
+            Modifier.semantics {
+                contentType = ContentType.NewPassword
+            },
+        )
+    }
+    // [END android_compose_autofill_4]
+}
+
+@Composable
+fun SaveDataWithAutofillOnClick() {
+    // [START android_compose_autofill_5]
+    val autofillManager = LocalAutofillManager.current
+    Column {
+        BasicTextField(
+            state = remember { TextFieldState() },
+            modifier =
+            Modifier.semantics {
+                contentType = ContentType.NewUsername
+            },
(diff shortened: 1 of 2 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -0,0 +1,138 @@
...
+            },
+        )
+    }
+    // [END android_compose_autofill_2]
+}
+
+@Composable
+fun SaveDataWithAutofill() {
+    // [START android_compose_autofill_3]
+    // [START android_compose_autofill_4]
```
mnuzen: We don't want this line here, right? This is only the third sample?

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -0,0 +1,138 @@
...
+    val autofillManager = LocalAutofillManager.current
```
mnuzen, line 72: Is this val used/needed in the sample?

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -0,0 +1,138 @@
...
+    Column {
+        BasicTextField(
+            state = remember { TextFieldState() },
+            modifier =
+            Modifier.semantics {
+                contentType = ContentType.NewUsername
+            },
+        )
+
+        Spacer(modifier = Modifier.height(16.dp))
```
mnuzen, line 91: For all following samples: not sure if the spacer is needed if it's just the snippet — but leaving this up to you

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -0,0 +1,138 @@
...
+// [START android_compose_autofill_6]
+@Composable
+fun customizeAutofillHighlight() {
+    val customHighlightColor = Color.Red
+    val usernameState = remember { TextFieldState() }
+
+    CompositionLocalProvider(LocalAutofillHighlightColor provides customHighlightColor) {
+        Column {
+            BasicTextField(
+                state = usernameState,
```
mnuzen: For consistency, should we just get rid of the `usernameState` val and replace inline on line 130 `state = remember ( TextFieldState() }` like the other samples do?

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -0,0 +1,138 @@
...
+import androidx.compose.ui.semantics.contentType
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.unit.dp
+import com.example.compose.snippets.touchinput.Button
+
+@Composable
+fun AddAutofill() {
+    // [START android_compose_autofill_1]
+    BasicTextField(
+        state = remember { TextFieldState("Enter your username.") },
```
mnuzen: I think we should either have initial text for all `TextFieldState`s or none. Leaving this up to you, but as long as we're consistent I'm fine with either.

mnuzen (changes requested):
I notice some other code was added in `AnimatedVisibilitySharedElementBlurSnippet.kt` and `UserInteractions.kt` — are these needed? Otherwise, generally looks good!

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
MagicalMeghan: i combined both into this one!

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
MagicalMeghan: yes

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
MagicalMeghan: not super needed but i like it there just in case they run it

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
mnuzen: Makes sense, thanks!

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
mnuzen: Leaving it SGTM!

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
MagicalMeghan: yeah i'll update

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
MagicalMeghan: removing for simplicity

TarheelCoupe43 (approved):

compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -0,0 +1,147 @@
...
+
+@Composable
+fun AddAutofill() {
+    var textFieldValue = remember {
+        mutableStateOf(TextFieldValue(""))
+    }
+    // [START android_compose_autofill_1]
+    TextField(
+        value = textFieldValue.value,
+        onValueChange = {textFieldValue.value = it},
```
StefMa, line 47:
```suggestion
        onValueChange = { textFieldValue.value = it },
```

gradle/libs.versions.toml
```diff
@@ -23,7 +23,7 @@ androidxHiltNavigationCompose = "1.2.0"
 coil = "2.7.0"
 # @keep
 compileSdk = "35"
-compose-latest = "1.7.6"
+compose-latest = "1.8.0-alpha08"
```
StefMa, line 26:
```suggestion
compose-latest = "1.8.0"
```

StefMa (commented):
Seems the PR is already used in the [Android docs](https://developer.android.com/develop/ui/compose/text/autofill).
I found especially the `{}` witout a space annoying 🫣 
Screenshot 2025-04-28 at 9 26 57 AM

PR #464, opened 2025-02-27
Add snippet showing system bar protection.

ashnohe:
Status bar has a gradient behind it so scrolling content doesn't clash with the status bar icons. Navigation bar is translucent so that buttons are visible as content scrolls behind.

Status bar and nav bar will look mostly like this (except navigation bar color will be a bit whiter):

https://github.com/user-attachments/assets/eb30909e-d681-484c-bdb2-38fd736bb6e9

Is this UI showing the scrolling list okay or should it be simplified?

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
@@ -80,9 +80,8 @@ private fun StatusBarProtection(
     heightProvider: () -> Float = calculateGradientHeight(),
 ) {
 
-    val calculatedHeight = heightProvider()
-
     Canvas(Modifier.fillMaxSize()) {
+        val calculatedHeight = heightProvider()
         val gradient = Brush.verticalGradient(
             colors = listOf(
                 color.copy(alpha = 1f),
--- a/compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
@@ -16,7 +16,6 @@
 
 package com.example.compose.snippets.layouts
 
-import android.os.Build
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
--- a/compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
@@ -115,7 +115,7 @@ fun MyContent() {
         LazyColumn(
             contentPadding = innerPadding
         ) {
-            items(13) {index ->
+            items(13) { index ->
                 Card(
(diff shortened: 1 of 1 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
```diff
@@ -0,0 +1,179 @@
...
+class SystemBarProtectionSnippets : ComponentActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        enableEdgeToEdge()
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            // Only impacts three-button navigation
+            window.isNavigationBarContrastEnforced = true
+        }
```
alexvanyo: Is this the default with `enableEdgeToEdge()`?
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:activity/activity/src/main/java/androidx/activity/EdgeToEdge.kt;l=325-326;drc=8a0d817ef0f9c48906a6466470163d301d91d024?q=enableEdgeToEdge

compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
```diff
@@ -0,0 +1,179 @@
...
+            brush = gradient,
+            size = Size(size.width, calculatedHeight),
+        )
+    }
+}
+
+@Composable
+fun calculateGradientHeight(): () -> Float {
+    val topInsets = WindowInsets.statusBars.getTop(LocalDensity.current).times(1.2F)
+    return { topInsets }
```
alexvanyo: By retrieving the `WindowInsets` and `Density` in composition and then doing the calculation in the lambda, any inset animations can be deferred to when the lambda is called instead of in composition:

```suggestion
    val statusBars = WindowInsets.statusBars
    val density = LocalDensity.current
    return { statusBars.getTop(density).times(1.2f) }
```

compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
```diff
@@ -0,0 +1,179 @@
...
+
+@Composable
+private fun StatusBarProtection(
+    color: Color = MaterialTheme.colorScheme.surfaceContainer,
+    heightProvider: () -> Float = calculateGradientHeight(),
+) {
+
+    val calculatedHeight = heightProvider()
+
+    Canvas(Modifier.fillMaxSize()) {
```
alexvanyo: Moving the `heightProvider` calcluation inside the drawing block for `Canvas` will also help defer reading the height until when it is needed:

```suggestion
    Canvas(Modifier.fillMaxSize()) {
        val calculatedHeight = heightProvider()
```

compose/snippets/src/main/java/com/example/compose/snippets/layouts/SystemBarProtectionSnippets.kt
ashnohe: yes, removed this and added an explanation about enableEdgeToEdge instead

PR #465, opened 2025-03-04
Adding Glance Inner Radius Snipper

secondsun:
Adding a snippet for Glance Inner Radius

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
@@ -39,6 +39,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.DpSize
 import androidx.compose.ui.unit.dp
 import androidx.glance.Button
@@ -69,6 +70,7 @@ import androidx.glance.appwidget.action.ActionCallback
 import androidx.glance.appwidget.action.actionRunCallback
 import androidx.glance.appwidget.action.actionSendBroadcast
 import androidx.glance.appwidget.action.actionStartService
+import androidx.glance.appwidget.cornerRadius
 import androidx.glance.appwidget.provideContent
 import androidx.glance.appwidget.updateAll
 import androidx.glance.appwidget.updateIf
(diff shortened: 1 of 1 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
```diff
@@ -868,6 +870,34 @@ object GlanceTheming {
...
+        if (Build.VERSION.SDK_INT < 31) {
+            return this
+        }
+
+        // Padding around the widget
+        val widgetPadding = 12.dp
+        val resources = LocalContext.current.resources
+        // get dimension in float (without rounding).
+        val px = resources.getDimension(android.R.dimen.system_app_widget_background_radius)
+        val widgetBackgroundRadiusDpValue = px / resources.displayMetrics.density
```
riggaroo, line 890: Can you use the helper function:
```
val widgetBackgroundRadiusDpValue =  with(LocalDensity.current){ intValue.toDp() }
```

compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
```diff
@@ -868,6 +870,34 @@ object GlanceTheming {
...
+     */
+    @Composable
+    fun GlanceModifier.appWidgetInnerCornerRadius(): GlanceModifier {
+        // [START android_compose_glance_innercornerradius]
+        if (Build.VERSION.SDK_INT < 31) {
+            return this
+        }
+
+        // Padding around the widget
+        val widgetPadding = 12.dp
```
riggaroo: Should this be passed in to the modifier?

compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
```diff
@@ -868,6 +870,34 @@ object GlanceTheming {
...
 
+object GlanceInnerPadding {
+
+    /**
+     * Applies corner radius for views that are visually positioned [widgetPadding]dp inside of the
+     * widget background.
+     */
+    @Composable
+    fun GlanceModifier.appWidgetInnerCornerRadius(): GlanceModifier {
+        // [START android_compose_glance_innercornerradius]
```
riggaroo: You may want to start this snippet with the @Composable part since you are returning something at the end of the function as part of the snippet, so it makes sense to see the full function definition.

compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
secondsun: LocalDensity isn't available during a glance composition.  This is a glance snippet so the code would crash.

compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
secondsun: Sounds good!

compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
secondsun: Updated!

riggaroo (approved):

PR #472, opened 2025-03-14
Add activity embedding snippets.

JonEckenrode:


gradle/libs.versions.toml
```diff
@@ -45,7 +49,7 @@ material3-adaptive = "1.1.0"
 material3-adaptive-navigation-suite = "1.3.1"
 media3 = "1.5.1"
 # @keep
-minSdk = "21"
+minSdk = "24"
```
riggaroo: This should probably not be bumped up so high, as it bumps up all samples.

riggaroo: if you need this for a specific sample, consider creating a new module that has this as the minimum requirement.

JonEckenrode: Restored minSdk to 21. Added @RequiresApi(api=VERSION_CODES.N) to enclosing method of snippets.

riggaroo (approved):

PR #489, opened 2025-04-04
Add restore credential snippets

neelanshsahai:


cy245 (approved):

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/RestoreCredentialsFunctions.kt
```diff
@@ -0,0 +1,54 @@
+package com.example.identity.credentialmanager
```
cy245, line 17: Please add license or run spotless to add it here

neelanshsahai: I dont think spotless is configured for this project. Should I add spotless configuration in this PR as well?

PR #494, opened 2025-04-09
Add smart lock migration snippets

neelanshsahai:


identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/SmartLockToCredMan.kt
```diff
@@ -0,0 +1,145 @@
...
+
+// [START android_identity_get_cred_request]
+val getCredRequest = GetCredentialRequest(
+    listOf(getPasswordOption)
+)
+// [END android_identity_get_cred_request]
+
+const val TAG: String = "tag"
+
+class LaunchSignInFlow(
```
cy245: Can we keep the class naming the same as the file? Also move the global variables into the class?

cy245 (approved):

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/SmartLockToCredMan.kt
cy245: Please resolve when addressed

PR #495, opened 2025-04-11
Add snippets for migration from Fido2 to Credman

neelanshsahai:


build.gradle.kts
```diff
@@ -4,7 +4,9 @@ plugins {
     alias(libs.plugins.version.catalog.update)
     alias(libs.plugins.android.application) apply false
     alias(libs.plugins.android.library) apply false
+    // [START android_identity_fido2_migration_dependency]
```
cy245: @sanbeiji the original code snippet includes the `plugins` outer layer to make it clearer where this should go - should we still include it? If so we'd need to duplicate it separately.

cy245: @neelanshsahai this also should go in the identity/credentialmanager directory.

identity/credentialmanager/build.gradle.kts
```diff
@@ -60,6 +60,8 @@ dependencies {
     implementation(libs.androidx.credentials.play.services.auth)
```
cy245, line 62: @neelanshsahai part of this section also needs to be referenced. Please mention this in the description; also pending @sanbeiji's guidance on if we should partially duplicate these sections.

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/Fido2ToCredmanMigration.kt
```diff
@@ -0,0 +1,239 @@
...
+    private val PUBLIC_KEY = ""
+
+    // [START android_identity_fido2_credman_init]
+    val credMan = CredentialManager.create(context)
+    // [END android_identity_fido2_credman_init]
+
+    // [START android_identity_fido2_migration_post_request_body]
+    suspend fun registerRequest() {
+        // ...
+        Builder()
```
cy245: @neelanshsahai Please add a comment that this is the okhttp3.Request.Builder since it's not clear from the snippet

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/Fido2ToCredmanMigration.kt
```diff
@@ -0,0 +1,239 @@
...
+            .url("$BASE_URL/<your api url>")
```
cy245: Is this mandatory?

build.gradle.kts
neelanshsahai: In documentation it was supposed to be a part of project level gradle file. Should I still move it to module level? If it only fetches the code snippet I think we can do it.

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/Fido2ToCredmanMigration.kt
neelanshsahai: the original snippet was this
```
suspend fun registerRequest(sessionId: String ... {
    // ...
    .method("POST", jsonRequestBody {
        name("attestation").value("none")
        name("authenticatorSelection").objectValue {
            name("residentKey").value("required")
        }
    }).build()
    // ...
}
```

Which was causing compile error. So added a builder class. I can make it similar to the other calls though

build.gradle.kts
cy245: Yes let's keep it at the module level - please leave a comment about the change.

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/Fido2ToCredmanMigration.kt
cy245: The Builder class is fine, but the url field isn't set in the original snippet

identity/credentialmanager/build.gradle.kts
neelanshsahai: Already added a comment
https://docs.google.com/spreadsheets/d/1G14c4PHQk082Bhb8INbsotFFQ5B2_ZDuNllc21lY4kM/edit?resourcekey=0-xwQDI-4byMjNVFOhnbATnQ&disco=AAABhrM_54g

We already have duplicate dependencies here

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/Fido2ToCredmanMigration.kt
neelanshsahai: Added a client.newCall like all other calls.

cy245 (approved):

PR #501, opened 2025-04-18
Add snippets for Credential Provider

neelanshsahai:


identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/CredentialProviderDummyActivity.kt
```diff
@@ -0,0 +1,491 @@
...
+    private fun handleUserSelectionForPasswordAuthentication(
+        mDatabase: MyDatabase,
+        callingAppInfo: CallingAppInfo,
+    ) {
+        // [START android_identity_credential_provider_user_selection_password]
+        val getRequest = PendingIntentHandler.retrieveProviderGetCredentialRequest(intent)
+
+        val passwordOption = getRequest?.credentialOptions?.first() as GetPasswordOption
+
+        val username = passwordOption.allowedUserIds.first()
```
neelanshsahai, line 376: Not really sure on this one. Might need a confirmation if this line is technically correct or not.
The original DAC page had this in the code snippet
```
val passwordOption = getRequest.credentialOption as GetPasswordCredentialOption

val username = passwordOption.username
```

cy245: I think this should be fine

cy245 (approved):

sanbeiji (approved):

PR #503, opened 2025-04-22
Add WebView snippets

cy245:
Added snippets for https://developer.android.com/identity/sign-in/credential-manager-webview. Some slight modifications:

- Combined first and second snippets since the former is a subset of the latter and it's awkward to break up a class declaration into 2 parts
- Updated "Initialize the WebView" section to include function header to make it clearer
- Combined last 2 snippets since they're part of the same function
- "INJECTED_VAL" is a very long string that is from the original app. While not part of a snippet, I've kept as-is for future reference

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -68,6 +68,7 @@ wearComposeMaterial = "1.4.1"
 wearToolingPreview = "1.0.0"
 activityKtx = "1.10.0"
 okHttp = "4.12.0"
+webkit = "1.13.0"
 
 [libraries]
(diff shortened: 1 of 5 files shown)


neelanshsahai:
LGTM
Nit: Spotless defaults to 4 spaces per tab indentation. Some files in the project have 2 space indent, and some have 4 space indent.
Adding spotless. We can run a quick build after merging this

neelanshsahai (approved):

cy245:
> LGTM Nit: Spotless defaults to 4 spaces per tab indentation. Some files in the project have 2 space indent, and some have 4 space indent. After adding spotless, we can run a quick build post merging this

Thanks Neelansh. Let's run a quick followup CL to fix the formatting w/Spotless since it affects other files too.

PR #512, opened 2025-05-08
Icon button snippets

wardlauren:
Adding icon button snippets.

Updating toml file to add dependency for extended icons.


JolandaVerhoef:
@riggaroo can you inform us whether the material extended icons are still okay to use or if we should use hard-coded drawables instead?

riggaroo:
We should be using hardcoded drawables created through Android Studio tooling - material-icons-extended is not the up-to-date set of icons that we should be recommending to developers to use.

compose/snippets/src/main/java/com/example/compose/snippets/components/IconButton.kt
```diff
@@ -0,0 +1,127 @@
...
+@Composable
+fun ToggleIconButtonExample() {
+    // isToggled initial value should be read from a view model or persistent storage.
+    var isToggled by rememberSaveable { mutableStateOf(false) }
+
+    IconButton(
+        onClick = { isToggled = !isToggled }
+    ) {
+        Icon(
+            imageVector = if (isToggled) Icons.Filled.Favorite else Icons.Outlined.FavoriteBorder,
```
riggaroo: This should also switch to XML files

riggaroo (approved):

JolandaVerhoef (approved):

Issue #179, opened 2024-01-03
Build fails for PRs opened from forks


alexvanyo:
https://github.com/android/snippets/blob/87948ac8ef0173a1562ed43d58801d76e8387f99/.github/workflows/build.yml#L31-L34

The checkout build step uses a secret PAT, which forks don't have access to, so the build step immediately fails for any PRs opened from a fork.

Example PR: https://github.com/android/snippets/pull/178

arriolac:
I made changes to fix the PAT-specific issue [here](https://github.com/android/snippets/commit/fdc1b8e2911e0bf98181f5b40e7101aa34ad395b). Builds are still failing though

wgobezie27:
Please any one help me this code development android app, because i am not exprianced to developed?
// MainActivity.kt
package com.example.constructionapp

import android.os.Bundle
import androidx.activity.ComponentActivity
import androidx.activity.compose.setContent
import androidx.compose.foundation.layout.*
import androidx.compose.material.*
import androidx.compose.material.icons.Icons
import androidx.compose.material.icons.filled.Add
import androidx.compose.runtime.*
import androidx.compose.ui.Modifier
import androidx.compose.ui.tooling.preview.Preview
import androidx.compose.ui.unit.dp
import androidx.compose.ui.Alignment
import androidx.navigation.NavHostController
import androidx.navigation.compose.NavHost
import androidx.navigation.compose.composable
import androidx.navigation.compose.rememberNavController
import com.example.constructionapp.ui.theme.ConstructionAppTheme
import androidx.compose.foundation.lazy.LazyColumn
import androidx.compose.foundation.lazy.items
import androidx.compose.runtime.saveable.rememberSaveable
import androidx.compose.foundation.text.KeyboardOptions
import androidx.compose.ui.text.input.KeyboardType
import androidx.compose.material3.AlertDialog
import androidx.compose.material3.Button
import androidx.compose.material3.ExperimentalMaterial3Api
import androidx.compose.material3.Icon
import androidx.compose.material3.IconButton
import androidx.compose.material3.MaterialTheme
import androidx.compose.material3.OutlinedTextField
import androidx.compose.material3.Scaffold
import androidx.compose.material3.Surface
import androidx.compose.material3.Text
import androidx.compose.material3.TopAppBar
import androidx.compose.runtime.saveable.rememberSaveable
import androidx.compose.ui.text.TextStyle
import androidx.compose.ui.unit.sp
import java.util.UUID
[...]

PR #513, opened 2025-05-08
Adding snippet for MV-HEVC video playback

vad710:


devbridie:
Due to #179 spotless will fail. It seems OK to me though.

PR #515, opened 2025-05-12
Fix package name

kul3r4:
Update package name to correclty nest inside existing Wear Package.


ithinkihaveacat (approved):
LGTM. Tested in Android Studio Meerkat Feature Drop | 2024.3.2 RC 4; both app and tile build and deploy as expected.

Issue #533, opened 2025-05-30
Bug with animated CircleNode modifier


IanGClifton:
There's a subtle bug with this code:

```
class CircleNode(var color: Color) : Modifier.Node(), DrawModifierNode {
    private val alpha = Animatable(1f)

    override fun ContentDrawScope.draw() {
        drawCircle(color = color, alpha = alpha.value)
        drawContent()
    }

    override fun onAttach() {
        coroutineScope.launch {
            alpha.animateTo(
                0f,
                infiniteRepeatable(tween(1000), RepeatMode.Reverse)
            ) {
            }
        }
    }
}
```

Alpha is initially set to 1f and animated toward 0f. This works as expected until the Node is reused. If you scroll inside a lazy list, the Node will be reused and its value will be somewhere between 0f and 1f, so let's say 0.5f. When it is attached, `alpha.animateTo` will now animated from 0f to 0.5f. If you continue to scroll, this node will keep getting reattached with a smaller and smaller animation alpha until the animation is effectively not noticeable (though still running).

PR #534, opened 2025-05-30
Fix for #533 animated CircleNode modifier

IanGClifton:
This addresses https://github.com/android/snippets/issues/533 by recreating the Animatable when the node is attached.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/modifiers/CustomModifierSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/modifiers/CustomModifierSnippets.kt
@@ -18,6 +18,7 @@ package com.example.compose.snippets.modifiers
 
 import android.annotation.SuppressLint
 import androidx.compose.animation.core.Animatable
+import androidx.compose.animation.core.AnimationVector1D
 import androidx.compose.animation.core.DecayAnimationSpec
 import androidx.compose.animation.core.RepeatMode
 import androidx.compose.animation.core.animateFloatAsState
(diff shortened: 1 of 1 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/modifiers/CustomModifierSnippets.kt
```diff
@@ -259,14 +260,15 @@ class ScrollableNode :
...
-        private val alpha = Animatable(1f)
+        private lateinit var alpha: Animatable<Float, AnimationVector1D>
 
         override fun ContentDrawScope.draw() {
             drawCircle(color = color, alpha = alpha.value)
             drawContent()
         }
 
         override fun onAttach() {
+            alpha = Animatable(1f)
```
mlykotom, line 271: I wonder whether it'd be more apropriate if we keep this 
`private var alpha = Animatable(1f)` but override `onReset()` and override with a new value. 
If someone at some poitn touches the `alpha` before `onAttach`, it won't crash the app.

mlykotom (approved):

compose/snippets/src/main/java/com/example/compose/snippets/modifiers/CustomModifierSnippets.kt
IanGClifton: Yeah, that could work too. For a simple/tiny example like this, it probably doesn't matter which way it's handled, but I could see that being a good option if the class is likely to grow and the lifecycle becomes harder to reason about. I'll leave it up to you fine folks to figure out which you prefer since the DAC docs probably have to get updated to explain either option and I don't have permissions to merge anyway 😆

mlykotom: Added a note on DAC. Thanks Ian! ❤️

PR #544, opened 2025-06-12
Update android_system_bar_protection_kotlin to change alpha

ashnohe:


views/src/main/java/insets/SystemBarProtectionSnippet.kt
```diff
@@ -51,18 +51,18 @@ class SystemBarProtectionSnippet : AppCompatActivity() {
             insets
         }
 
-        // [START android_system_bar_protection_kotlin]
         val red = 52
         val green = 168
         val blue = 83
+        val paneBackgroundColor = Color.argb(255, red, green, blue)
```
alabiaga: if we are making it completely opaque just use  https://developer.android.com/reference/android/graphics/Color#rgb(int,%20int,%20int)

alabiaga (approved):

PR #546, opened 2025-06-17
Fix code to also support external apps

tiwiz:


compose/snippets/src/main/java/com/example/compose/snippets/draganddrop/DragAndDropSnippets.kt
```diff
@@ -71,15 +73,30 @@ private fun DragAndDropSnippet() {
     }
     // [END android_compose_drag_and_drop_4]
 
+    val activity = LocalActivity.current
```
JolandaVerhoef: consider using `LocalActivity.current?.let { activity ->` so you have a non-null activity

compose/snippets/src/main/java/com/example/compose/snippets/draganddrop/DragAndDropSnippets.kt
```diff
@@ -71,15 +73,30 @@ private fun DragAndDropSnippet() {
...
+
     // [START android_compose_drag_and_drop_5]
+    val externalAppCallback = remember {
```
JolandaVerhoef: What are you going to do with `externalAppCallback?`

tiwiz: Not much, really, I just need it to update the guide [here](https://developer.android.com/develop/ui/compose/touch-input/user-interactions/drag-and-drop#receive-drop). Should we use it in the snippet as well?

compose/snippets/src/main/java/com/example/compose/snippets/draganddrop/DragAndDropSnippets.kt
tiwiz: Would it be clear, since it's outside the scope of the snippet?

JolandaVerhoef: I think it's okay. If people see you use `activity` they will know they need to get an activity from somewhere.

PR #549, opened 2025-06-19
Add snippets for Android Wear's Always On doc

ithinkihaveacat:
Will be used here: https://developer.android.com/training/wearables/always-on

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnActivity.kt
+++ b/wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnActivity.kt
@@ -105,7 +105,7 @@ fun ElapsedTime(ambientState: AmbientState) {
 
     val elapsedMs by
     produceState(initialValue = 0L, key1 = startTimeMs) {
-        while (true) {
+        while (true) { // time doesn't stop!
             value = SystemClock.elapsedRealtime() - startTimeMs
             // In ambient mode, update every minute instead of every second
             val updateInterval = if (ambientState.isAmbient) 60_000L else 1_000L


wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnActivity.kt
```diff
@@ -0,0 +1,188 @@
...
+
+@Composable
+// [START android_wear_ongoing_activity_elapsedtime]
+fun ElapsedTime(ambientState: AmbientState) {
+    // [START_EXCLUDE]
+    val startTimeMs = rememberSaveable { SystemClock.elapsedRealtime() }
+
+    val elapsedMs by
+    produceState(initialValue = 0L, key1 = startTimeMs) {
+        while (true) {
```
kul3r4: do you want to add a comment here on why you need while(true) -> it may seems strange having a non ending cycle?

wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnActivity.kt
```diff
@@ -0,0 +1,188 @@
...
+                    // [START_EXCLUDE]
+                    Spacer(modifier = Modifier.height(8.dp))
+                    SwitchButton(
+                        checked = isOngoingActivity,
+                        onCheckedChange = { newState ->
+                            Log.d("AlwaysOnActivity", "Switch button changed: $newState")
+                            isOngoingActivity = newState
+
+                            if (newState) {
+                                Log.d("AlwaysOnActivity", "Starting AlwaysOnService")
```
kul3r4: Nit: you can have TAG here as you do in other files

wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnActivity.kt
ithinkihaveacat: Done!

wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnActivity.kt
ithinkihaveacat: Thanks, done!

PR #550, opened 2025-06-20
Improve formatting of embedded snippet

ithinkihaveacat:
"// ..." is generated automatically by [START_EXCLUDE]/[END_EXCLUDE]


kul3r4:
as per chat: an alternative is to add something meaningful here explaining the code that we are excluding

PR #560, opened 2025-07-03
Add snippets for focus management in Compose

chikoski:
This pull request add some snippets about focus management


chikoski:
Let me update and come back to you.

PR #564, opened 2025-07-11
Add profiling manager snippets for how-to-capture DAC docs

edgararriagag:
Adding some snippets required for the upcoming Profiling Manager sections for DAC documents

### DIFF
--- a/misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
+++ b/misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
@@ -13,7 +13,6 @@ import androidx.core.os.Profiling;
 import androidx.core.os.SystemTraceRequestBuilder;
 import androidx.core.os.BufferFillPolicy;
 
-// [START android_profiling_manager_record_system_trace_java]
 
 public class MainActivityJava extends Activity {
 
@@ -23,6 +22,7 @@ public class MainActivityJava extends Activity {
     sampleRecordSystemTrace();
   }
 
+  // [START android_profiling_manager_record_system_trace_java]
   void heavyOperation() {
     // Computations you want to profile
   }
@@ -55,6 +55,7 @@ public class MainActivityJava extends Activity {
     requestBuilder.setTag("FOO");
     requestBuilder.setDurationMs(60000);
     requestBuilder.setBufferFillPolicy(BufferFillPolicy.RING_BUFFER);
+    requestBuilder.setBufferSizeKb(20971520);
     Profiling.requestProfiling(getApplicationContext(), requestBuilder.build(), mainExecutor,
         resultCallback);
 
@@ -67,6 +68,5 @@ public class MainActivityJava extends Activity {
     // Once the interesting code section is profiled, stop profile
     stopSignal.cancel();
   }
+  // [END android_profiling_manager_record_system_trace_java]
 }
-
-// [END android_profiling_manager_record_system_trace_java]
--- a/misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
+++ b/misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
@@ -1,0 +1,72 @@
+package com.example.snippets.profiling;
+
+import android.app.Activity;
+import android.os.Bundle;
+import android.util.Log;
+import java.util.function.Consumer;
+import java.util.concurrent.Executor;
+import android.os.ProfilingResult;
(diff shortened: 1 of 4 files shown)


misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
```diff
@@ -0,0 +1,72 @@
...
+  public void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    sampleRecordSystemTrace();
+  }
+
+  void heavyOperation() {
+    // Computations you want to profile
+  }
+
+  void sampleRecordSystemTrace() {
```
AliceYuan, line 30: should we just start the code snippet right before this function?

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.asExecutor
+
+// [START android_profiling_manager_record_system_trace_kotlin]
+
+class MainActivity : Activity() {
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    sampleRecordSystemTrace()
+  }
+
```
AliceYuan, line 24: same q around just showing the function

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
```diff
@@ -0,0 +1,72 @@
...
+                      + profilingResult.getErrorMessage());
+            }
+          }
+        };
+    CancellationSignal stopSignal = new CancellationSignal();
+
+    SystemTraceRequestBuilder requestBuilder = new SystemTraceRequestBuilder();
+    requestBuilder.setCancellationSignal(stopSignal);
+    requestBuilder.setTag("FOO");
+    requestBuilder.setDurationMs(60000);
```
AliceYuan, line 56: can we pull this into duration const

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.kt
```diff
@@ -0,0 +1,67 @@
...
+        )
+      }
+    }
+    val stopSignal = CancellationSignal()
+
+    val requestBuilder = SystemTraceRequestBuilder()
+    requestBuilder.setCancellationSignal(stopSignal)
+    requestBuilder.setTag("FOO") // Caller supplied tag for identification
+    requestBuilder.setDurationMs(60000)
+    requestBuilder.setBufferFillPolicy(BufferFillPolicy.RING_BUFFER)
```
AliceYuan, line 48: can we share what the other options are in a comment?

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
edgararriagag: Is there a reason to want to pull this into a const, if we could, I prefer to leave as is because that way we make the snippet as hermatic as possible and we don't have to tweak the current wording of the snippet explanation in the DAC docs.

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.kt
edgararriagag: The constructor methods are subject to change, we did not mean to be exhaustive here but provide an example. I think there's only one more method not shown here which is the setBufferSizeKb, I added it instead to the sample itself, I think its a bit odd to add it as a comment, so instead I integrated it directly

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.kt
edgararriagag: Done

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerSnippets.java
edgararriagag: Done, I also did the same for kotlin snippet. ptal and reopen if I missed anything

bentrengrove (approved):

edgararriagag:
It looks like the bringing patches from forked repos doesn't work, so I will go ahead and post it directly into the main repository, the new PR is https://github.com/android/snippets/pull/568 . Closing this one as the review will finish in the new one.

PR #572, opened 2025-07-30
Add example of Lottie animation

ithinkihaveacat:
The snippet will go on <https://developer.android.com/training/wearables/tiles/animations>.

The `lottie.json` is a copy of https://github.com/google/lottie-tools/blob/main/tests/spec/valid/shape_transforms_animated.json.


ithinkihaveacat:
FYI @NedaTop.

PR #574, opened 2025-07-31
Add PreloadManager snippets, and update media3 to 1.8.0

asolovay:
Submitting doc snippets used for the new [PreloadManager documentation][plm-docs]. Had to bump the Media3 version up to 1.8.0 for some of the new APIs in that doc.

Once we've submitted this in Github, I'll update my docs to pull the code snippets in from there.

[plm-docs]: https://developer.android.com/media/media3/exoplayer/preloading-media/preloadmanager/


gradle/libs.versions.toml
```diff
@@ -187,6 +189,8 @@ play-services-wearable = { module = "com.google.android.gms:play-services-wearab
 validator-push = { module = "com.google.android.wearable.watchface.validator:validator-push", version.ref = "validatorPush" }
 wear-compose-material = { module = "androidx.wear.compose:compose-material", version.ref = "wearComposeMaterial" }
 wear-compose-material3 = { module = "androidx.wear.compose:compose-material3", version.ref = "wearComposeMaterial3" }
+androidx-media3-common-ktx = { group = "androidx.media3", name = "media3-common-ktx", version.ref = "media3CommonKtx" }
+media3-exoplayer = { group = "androidx.media3", name = "media3-exoplayer", version.ref = "media3Exoplayer" }
```
MayuriKhinvasara: Can we keep naming convention as same ? Eg change to "androidx-media3-exoplayer"

MayuriKhinvasara (commented):
Added one minor comment. LGTM otherwise

misc/src/main/java/com/example/snippets/PreloadManagerKotlinSnippets.kt
```diff
@@ -0,0 +1,134 @@
...
+        private fun fetchMedia(
+            preloadManager: DefaultPreloadManager,
+            mediaItem: MediaItem,
+            player: ExoPlayer,
+            currentIndex: Int
+        ) {
+            // [START android_defaultpreloadmanager_getAndPlayMedia]
+            // When a media item is about to display on the screen
+            val mediaSource = preloadManager.getMediaSource(mediaItem)
+            if (mediaSource != null) {
```
tianyif, line 101: Shall we change the L101-103 to 
```
if (mediaSource != null) {
    player.setMediaSource(mediaSource)
} else {
    // If the mediaItem hasn't been added to the preload manager yet,
    // set it directly to the player when it's about to play
    player.setMediaItem(mediaItem)
}
```

gradle/libs.versions.toml
asolovay: Thanks for spotting that! I'd just accepted Android Studio's proposed fix, but in fact, we already include Media3 common & exoplayer at lines 136-7 -- I just needed to update the version number (to 1.8.0) and add those labels to the dependencies. PTAL!

misc/src/main/java/com/example/snippets/PreloadManagerKotlinSnippets.kt
asolovay: Thanks! Yes, that was a quick cleanup just to get it to compile -- this version makes it a better example.

I changed the comment slightly, PTAL.

MayuriKhinvasara (approved):

PR #575, opened 2025-08-04
Fixing compile error in DeviceCompatibilityModeTestJavaSnippets

asolovay:
Happened to notice a typo in the file, but it still didn't compile when I fixed the typo, so I made a second fix.

The typo was a missing semicolon at the "import AppCompatActivity" line. But after I fixed that, I got an error complaining about `assertFalse(isLetterboxed(activity))` because `isLetterboxed` is expecting an AppCompatActivity but `activity` is a `MainActivity`. Changing `isLetterboxed` fixed the compile error and also meant we don't import AppCompatActivity so the original typo goes away. [shrug emoji]

Is this fix okay?

### DIFF
--- a/misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java
+++ b/misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java
@@ -16,7 +16,6 @@
 
 package com.example.snippets;
 
-import androidx.appcompat.app.AppCompatActivity
 import androidx.test.core.app.ActivityScenario;
 import androidx.test.ext.junit.rules.ActivityScenarioRule;
 import org.junit.Rule;
@@ -42,7 +41,7 @@ public class DeviceCompatibilityModeTestJavaSnippets {
 
 
     // Method used by snippets.
-    public boolean isLetterboxed(AppCompatActivity activity) {
+    public boolean isLetterboxed(MainActivity activity) {
         return true;
     }
 


JonEckenrode:
What was the typo, Andrew? I don't see where it was fixed. Also, these files compiled previously.

asolovay:
The typo was in line 19:
https://github.com/android/snippets/blob/main/misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java#L19

```
import androidx.appcompat.app.AppCompatActivity
import androidx.test.core.app.ActivityScenario;
import androidx.test.ext.junit.rules.ActivityScenarioRule;
```

(note the lack of a semicolon at the end of the first line).

But when I fixed that typo, the file still wouldn't compile, because isLetterBoxed() expects to be passed an AppCompatActivity but it's actually being passed a MainActivity.

When I changed the isLetterBoxed method to be passed a MainActivity, everything compiles -- but since the file no longer needs AppCompatActivity, AndroidStudio removed that "import AppCompatActivity" line, so the typo fix went away.

...just to confirm, when you submitted it, did you compile the AndroidTest module? I don't see how it could have compiled with that missing semicolon.

misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java
```diff
@@ -42,7 +41,7 @@ public void activity_launched_notLetterBoxed() {
 
 
     // Method used by snippets.
-    public boolean isLetterboxed(AppCompatActivity activity) {
+    public boolean isLetterboxed(MainActivity activity) {
```
florina-muntenescu, line 44: Why did you move to MainActivity instead of AppCompatActivity? 
If the problem was a `;` missing, this shouldn't be needed.

asolovay: As mentioned above -- when I added the semicolon I still couldn't compile, because isLetterBoxed() 's return value was being passed to something that expected a MainActivity. I don't know if this was the best way to fix that but it works and it doesn't change the code that's actually included in the docs, so 🤷

Thanks for submitting!

PR #579, opened 2025-08-05
Revised snippets for consistency.

JonEckenrode:
Revised snippets for consistency and correctness.

### DIFF
--- a/misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java
+++ b/misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java
@@ -16,6 +16,7 @@
 
 package com.example.snippets;
 
+import androidx.appcompat.app.AppCompatActivity;
 import androidx.test.core.app.ActivityScenario;
 import androidx.test.ext.junit.rules.ActivityScenarioRule;
 import org.junit.Rule;
@@ -40,9 +41,11 @@ public class DeviceCompatibilityModeTestJavaSnippets {
     // [END android_device_compatibility_mode_assert_isLetterboxed_java]
 
 
-    // Method used by snippets.
-    public boolean isLetterboxed(MainActivity activity) {
-        return true;
-    }
+    // Class used by snippets.
 
+    class MainActivity extends AppCompatActivity {
+        public boolean isLetterboxed (MainActivity activity) {
+            return true;
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


asolovay:
Did you try compiling this code? I just tried on my machine and I got an error:

```
/usr/local/google/home/asolovay/AndroidStudioProjects/snippets/misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java:37: error: cannot find symbol
                    assertFalse(isLetterboxed(activity));
                                ^
  symbol:   method isLetterboxed(DeviceCompatibilityModeTestJavaSnippets.MainActivity)
  location: class DeviceCompatibilityModeTestJavaSnippets
  ```
  
  I *think* that's because `isLetterboxed()` is now an internal method in the new MainActivity class.
  
Just a reminder, the checks don't currently run against code in androidTest, so you can't count on Github to tell you the code compiles.

**Comment by JonEckenrode** [comment]
Not sure what you mean, Andrew, The class that defines is letterboxed() is a helper class. It's not a snippet that's pulled from GitHub. But that class is added by this PR. There was just a method before.

File: misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java [hunk 1]
```diff
@@ -33,16 +34,18 @@ public void activity_launched_notLetterBoxed() {
...
-    public boolean isLetterboxed(MainActivity activity) {
-        return true;
-    }
+    // Class used by snippets.
 
+    static class MainActivity extends AppCompatActivity {
+        public boolean isLetterboxed (MainActivity activity) {
+            return true;
+        }
+    }
```
alexvanyo: The usage shouldn't need to refer to the `activity` twice - instead of having the class be static, and take a `MainActivity` as an argument, it should be able to mirror the Kotlin snippet directly:

```suggestion
    class MainActivity extends AppCompatActivity {
        public boolean isLetterboxed() {
            return true;
        }
    }
```

misc/src/androidTest/java/com/example/snippets/DeviceCompatibilityModeTestJavaSnippets.java
```diff
@@ -33,16 +34,18 @@ public void activity_launched_notLetterBoxed() {
         try (ActivityScenario<MainActivity> scenario =
             ActivityScenario.launch(MainActivity.class)) {
                 scenario.onActivity( activity -> {
-                    assertFalse(isLetterboxed(activity));
+                    assertFalse(activity.isLetterboxed(activity));
```
alexvanyo: There is no parameter now to this method, so we can remove the `activity` being passed here.

```suggestion
                    assertFalse(activity.isLetterboxed());
```

alexvanyo (approved):

PR #582, opened 2025-08-07
Adding code snippets for wake locks doc

asolovay:
Starting with the code snippets to add and release a wake lock, we can add other snippets later.

Used by docs in:

https://developer.android.com/develop/background-work/background-tasks/awake/wakelock/

The actual snippets used on the DAC pages are bracketed by the `[START ...]` and `[END ...]` comments; everything else is just scaffolding to get the snippet to compile.


misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsJava.java
```diff
@@ -0,0 +1,38 @@
...
+        PowerManager powerManager = (PowerManager) getSystemService(POWER_SERVICE);
+        PowerManager.WakeLock wakeLock = powerManager.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "MyClassName::MyWakelockTag");
+        wakeLock.acquire();
+        // [END android_backgroundwork_wakelock_create_java]
+
+        super.onCreate(savedInstanceState);
+    }
+
+    // [START android_backgroundwork_wakelock_release_java]
+    void doSomethingAndRelease() throws MyException {
```
AliceYuan, line 26: what if it just throws `Exception` instead? I'm not sure if we need the custom exception here

PR #584, opened 2025-08-08
Resolve build failures so repository will build with `./gradlew clean build`

cartland:
Key changes include:

   * Permission Handling:
       * Add @RequiresPermission annotation to methods in the `bluetoothle` sample.
       * Added the ACCESS_COARSE_LOCATION permission to the `bluetoothle` manifest to comply with Android 12+ requirements for apps requesting ACCESS_FINE_LOCATION.
   * Manifest Fixes:
       * Corrected an invalid icon placeholder in the `identity/credentialmanager` manifest that was causing a build error.
       * Removed the deprecated package attribute from the `bluetoothle` manifest.


cartland:
I could not reproduce this failure https://github.com/android/snippets/actions/runs/16839790873/job/47708049483.

I added --stacktrace to the build workflow to understand why it is failing.

cartland:
We could probably simplify the build script. Our previous GitHub Actions workflow was missing some modules, so we could choose to run `./gradlew build` instead of separate build commands for each module.

```
    - name: Build All
      run: ./gradlew build --stacktrace
    - name: Build Watch Face Push validation snippets
      run: ./gradlew :watchfacepush:validator:run --stacktrace
```

bentrengrove:
Nice! Let's change the build script in this PR as well and see if it passes

.github/workflows/build.yml
```diff
@@ -38,16 +38,18 @@ jobs:
...
+      run: ./gradlew :kotlin:build --stacktrace
     - name: Build Wear snippets
-      run: ./gradlew :wear:build
+      run: ./gradlew :wear:build --stacktrace
     - name: Build misc snippets
-      run: ./gradlew :misc:build
+      run: ./gradlew :misc:build --stacktrace
     - name: Build XR snippets
-      run: ./gradlew :xr:build
+      run: ./gradlew :xr:build --stacktrace
```
riggaroo: We can probably remove this if the build is running again at the end

cartland:
@riggaroo @bentrengrove I simplified the build script to `./gradlew build` so we don't need to specify each module. Checks passed. Let me know if you have any other suggestions or you can merge.

PR #598, opened 2025-08-22
Wear complications

garanj:
Adds some more snippets for complications


wear/src/main/java/com/example/wear/snippets/complication/ConfigurationActivity.kt
```diff
@@ -0,0 +1,66 @@
+package com.example.wear.snippets.complication
```
kul3r4, line 17: is this missing copyright?

kul3r4 (commented):
Only one comment about missing Copyright

ithinkihaveacat (approved):

wear/src/main/java/com/example/wear/snippets/complication/ConfigurationActivity.kt
garanj: Oh! Sorry!! Added

PR #609, opened 2025-08-29
Enforce ktlint max-line-length with Spotless gradle/init.gradle.kts

cartland:
Snippets from this repository are included in documentation at developer.android.com. Long lines are difficult to read in these snippets (and in general). We previously disabled `ktlint` for `max-line-length` because it was difficult to enforce across all snippets. Now we're seeing the challenge appear in the documentation publishing process. Since the snippet line length is not enforced in the code base, teams are adding manual warnings and checks at publishing time to make sure that the line length is not too long.

I am creating this pull request to show which changes will allow `max-line-length` to be enforced. I will ask teams to review these changes to make sure they don't cause other side effects in the documentation or build process. After everyone is satisfied, we can enforce the line lengths.

### DIFF
--- a/xr/src/main/java/com/example/xr/scenecore/SpatialVideo.kt
+++ b/xr/src/main/java/com/example/xr/scenecore/SpatialVideo.kt
@@ -126,7 +126,8 @@ private fun ComponentActivity.surfaceEntityCreateDRM(xrSession: Session) {
         stereoMode = SurfaceEntity.StereoMode.STEREO_MODE_SIDE_BY_SIDE,
         pose = Pose(Vector3(0.0f, 0.0f, -1.5f)),
         shape = SurfaceEntity.Shape.Quad(FloatSize2d(1.0f, 1.0f)),
-        surfaceProtection = SurfaceEntity.SurfaceProtection.SURFACE_PROTECTION_PROTECTED
+        surfaceProtection =
+        SurfaceEntity.SurfaceProtection.SURFACE_PROTECTION_PROTECTED
     )
 
     // Build a MediaItem with the necessary DRM configuration.
@@ -157,7 +158,8 @@ private fun ComponentActivity.surfaceEntityHDR(xrSession: Session) {
     // to your content.
     val hdrMetadata = SurfaceEntity.ContentColorMetadata(
         colorSpace = SurfaceEntity.ContentColorMetadata.ColorSpace.COLOR_SPACE_BT2020,
-        colorTransfer = SurfaceEntity.ContentColorMetadata.ColorTransfer.COLOR_TRANSFER_ST2084, // PQ
+        // PQ
+        colorTransfer = SurfaceEntity.ContentColorMetadata.ColorTransfer.COLOR_TRANSFER_ST2084,
         colorRange = SurfaceEntity.ContentColorMetadata.ColorRange.COLOR_RANGE_LIMITED,
         maxContentLightLevel = 1000 // Example: 1000 nits
     )
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -67,7 +67,6 @@ allprojects {
                 "standard:function-naming",
                 "standard:property-naming",
                 "standard:class-naming",
-                "standard:max-line-length",
                 "standard:comment-wrapping",
                 "standard:import-ordering",
                 "standard:filename",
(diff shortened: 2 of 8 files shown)


devbridie:
I am supportive of this, but I don't see how the modification of the lint rules follow the changes that are suggested in the PR - are there some changes that are manual instead of generated by the linter? Examples on the commits.

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/AnimatedVisibilitySharedElementSnippets.kt
```diff
@@ -211,22 +221,22 @@ fun SnackContents(
...
-            contentDescription = null
+            contentDescription = null,
         )
         Text(
             text = snack.name,
             modifier = Modifier
                 .wrapContentWidth()
                 .padding(8.dp),
-            style = MaterialTheme.typography.titleSmall
+            style = MaterialTheme.typography.titleSmall,
```
devbridie, line 239: Added trailing comma

compose/snippets/src/main/java/com/example/compose/snippets/animations/sharedelement/CustomizeSharedElementsSnippets.kt
```diff
@@ -603,9 +631,9 @@ fun PlaceholderSizeAnimated_Demo() {
                 }
             }
             composable(
-                "details/{id}",
+                route = "details/{id}",
```
devbridie, line 634: Named parameter added

compose/snippets/src/main/java/com/example/compose/snippets/components/IconButton.kt
```diff
@@ -85,7 +93,11 @@ fun MomentaryIconButton(
         interactionSource = interactionSource
     ) {
         Icon(
-            painter = if (isPressed) painterResource(id = selectedImage) else painterResource(id = unselectedImage),
+            painter = if (isPressed) {
```
devbridie, line 96: Brace style fixed

xr/src/main/java/com/example/xr/scenecore/SpatialVideo.kt
```diff
@@ -126,7 +126,8 @@ private fun ComponentActivity.surfaceEntityCreateDRM(xrSession: Session) {
         stereoMode = SurfaceEntity.StereoMode.STEREO_MODE_SIDE_BY_SIDE,
         pose = Pose(Vector3(0.0f, 0.0f, -1.5f)),
         shape = SurfaceEntity.Shape.Quad(FloatSize2d(1.0f, 1.0f)),
-        surfaceProtection = SurfaceEntity.SurfaceProtection.SURFACE_PROTECTION_PROTECTED
+        surfaceProtection =
+        SurfaceEntity.SurfaceProtection.SURFACE_PROTECTION_PROTECTED
```
devbridie, line 130: Is this correct? I would expect an indent here.

cartland: I would also expect an indent here. I'm not sure why Spotless or Android Studio suggested this formatting. If I have time, I can check to see if Spotless will let me add the indentation. I would normally expect Spotless to fix this, so maybe some of the other lint rules can be improved.

cartland:
Whenever possible, the changes come from `./gradlew spotlessApply`. When Spotless failed, I tried to make minimal changes and then use Android Studio to fix formatting. Sometimes these disagree with each other, so I did the bare minimum to make sure that `./gradlew spotlessCheck` passed.

In general, I think it's better to try and simplify the Spotless configuration so that it uses more and more of the default Android Studio style guides. Currently there are a lot of disabled rules, and I'm not sure how long those disabled rules interact with each other. I'm also not sure why there are two different line length rules that were disabled. 

Overall, I agree with your confusion. If we can fix more of these style issues with Spotless configuration, then we can spend less time discussing the indentation, named parameters, and trailing commas. 

The primary purpose of this change is to provide a step towards a more strict and standard Spotless configuration. If it's easier, you are welcome to close this pull request and create a new one with additional changes that go beyond this step. You are also welcome to merge this version and then create additional changes that build on it.

kkuan2011:
Discussed with Cartland about how we can move forward on this. I think we need to enable more of the Spotless rules one-by-one, and then hopefully enabling this rule will only change the line lengths and not change other parts of the code. So we don't need to enforce line length later down the pipeline. Thanks Cartland! We'll take a closer look at how to get to this state.

PR #610, opened 2025-08-29
Disabling some "unused" warnings

asolovay:
I'd submitted code that's never actually used in the android/snippets apps. Ideally we'd include it in an app, but in the meantime, disabling the "unused" warnings.

Also removed one line (outside of a published-on-DAC code block) that turned out not to be needed.

No changes to the code that's included on DAC, and the code still compiles.

### DIFF
--- a/misc/src/main/java/com/example/snippets/PreloadManagerKotlinSnippets.kt
+++ b/misc/src/main/java/com/example/snippets/PreloadManagerKotlinSnippets.kt
@@ -30,6 +30,7 @@ import java.lang.Math.abs
 // constants to make the code snippets work
 const val currentPlayingIndex = 10
 
+@Suppress("Unused")
 @UnstableApi
 // [START android_defaultpreloadmanager_MyTargetPreloadStatusControl]
 class MyTargetPreloadStatusControl(
@@ -57,6 +58,7 @@ class MyTargetPreloadStatusControl(
 }
 // [END android_defaultpreloadmanager_MyTargetPreloadStatusControl]
 
+@Suppress("Unused")
 class PreloadManagerSnippetsKotlin {
 
     class PreloadSnippetsActivity : AppCompatActivity() {
@@ -72,8 +74,6 @@ class PreloadManagerSnippetsKotlin {
                 DefaultPreloadManager.Builder(context, targetPreloadStatusControl)
             val preloadManager = preloadManagerBuilder.build()
             // [END android_defaultpreloadmanager_createPLM]
-
-            val player = preloadManagerBuilder.buildExoPlayer()
 
             // [START android_defaultpreloadmanager_addMedia]
             val initialMediaItems = pullMediaItemsFromService(/* count= */ 20)
--- a/misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsJava.java
+++ b/misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsJava.java
@@ -6,6 +6,7 @@ import android.os.PowerManager;
 
 import androidx.annotation.Nullable;
 
+@SuppressWarnings("Unused")
 public class WakeLockSnippetsJava extends Activity {
 
     PowerManager.WakeLock wakeLock;
@@ -23,6 +24,7 @@ public class WakeLockSnippetsJava extends Activity {
         super.onCreate(savedInstanceState);
     }
 
+    @SuppressWarnings("Unused")
     // [START android_backgroundwork_wakelock_release_java]
     void doSomethingAndRelease() throws MyException {
         try {
--- a/misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsKotlin.kt
+++ b/misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsKotlin.kt
@@ -22,6 +22,7 @@ import android.os.Bundle
 import android.os.PowerManager
 
 // Snippets for doc page go here
+@Suppress("Unused")
 class WakeLockSnippetsKotlin : Activity() {
 
     // [START android_backgroundwork_wakelock_create_kotlin]


misc/src/main/java/com/example/snippets/PreloadManagerKotlinSnippets.kt
```diff
@@ -30,6 +30,7 @@ import java.lang.Math.abs
 // constants to make the code snippets work
 const val currentPlayingIndex = 10
 
+@Suppress("Unused")
```
cartland: You can either use `@Suppress("UNUSED")` or `@Suppress("UNUSED_PARAMETER")`. If you only got a warning for the parameter, try to use the more specific option. Otherwise "UNUSED" is ok.

We should be consistent on capitalization, even though it acts case insensitive. All examples I see are either completely lowercase or completely uppercase. I am leaning towards uppercase.

misc/src/main/java/com/example/snippets/PreloadManagerKotlinSnippets.kt
```diff
@@ -57,6 +58,7 @@ class MyTargetPreloadStatusControl(
 }
 // [END android_defaultpreloadmanager_MyTargetPreloadStatusControl]
 
+@Suppress("Unused")
```
cartland: If we go with uppercase, this would be:
`@Suppress("UNUSED")`

Update the others.

misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsJava.java
```diff
@@ -6,6 +6,7 @@
 
 import androidx.annotation.Nullable;
 
+@SuppressWarnings("Unused")
```
cartland: Java all caps:
`@SuppressWarnings("UNUSED")`

misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsKotlin.kt
```diff
@@ -22,6 +22,7 @@ import android.os.Bundle
 import android.os.PowerManager
 
 // Snippets for doc page go here
+@Suppress("Unused")
```
cartland: `@Suppress("UNUSED")`

cartland (changes requested):
Check if you can use `UNUSED_PARAMETER` instead of `UNUSED` in one place. I also think we should use all capital letters.

misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsJava.java
asolovay: Thanks! Just wanted to confirm here that it's `@SuppressWarnings()` in Java, but `@Suppress()` in Kotlin?

asolovay: Actually -- when I amend my commit, I still get the warning "Method 'doSomethingAndRelease()' is never used" for this Java file (but *not* for the Kotlin file).

Is there something I'm doing wrong with `@SuppressWarnings()`? Right now I have:

```java
    @SuppressWarnings("UNUSED")
    // [START android_backgroundwork_wakelock_release_java]
    void doSomethingAndRelease() throws MyException {
        try {
            wakeLock.acquire();
            doTheWork();
        } finally {
            wakeLock.release();
        }
    }
    // [END android_backgroundwork_wakelock_release_java]

asolovay:
Cartland -- I made the updates you suggested, but I still seem to get a warning in `WakeLockSnippetsJava.doSomethingAndRelease()`. Can you see what I'm doing wrong there? Thanks much!

misc/src/main/java/com/example/snippets/backgroundwork/WakeLockSnippetsJava.java
cartland: That looks reasonable to me. Can you clarify where you are seeing the morning? Is this in Android Studio or a when you run a command line lint command? I can try your steps to reproduce.

asolovay: It's in Android Studio, when I try to push the commit to GitHub--it pops up a screen saying "there are these warnings, are you sure you want to push?"

But when I try compiling the "misc" module, I get no warnings. So if the answer is "pay attention to that, don't worry about the warning you get when you push", I'm happy.

cartland: It looks like we need to use `@SuppressWarnings("unused")` in Java. Android Studio lint only respects this if the "unused" is lowercase.

I'm thinking we should use lower case for both Kotlin: `@Suppress("unused")` and Java `@SuppressWarnings("unused")`. The capitalized version only work in Kotlin, so I would prefer to be consistent across languages by using lowercase.

cartland:
I think we need to use the lowercase version of Kotlin `Suppress("unused")` and Java `@SuppressWarnings("unused")`. In Java, Android Studio will help you add the Java annotation if you click Code -> Inspect Code then find a warning and click "Suppress". The Kotlin annotation will add a `/** @noinspection unused*/` comment but I think the annotation `Suppress("unused")` is better than the comment. However, the Android Studio "Suppress" tool is probably the easiest way to find the name of the warning that you want to suppress, since this should work for other types of warnings (not just unused code).

Suppress

asolovay:
> I think we need to use the lowercase version of Kotlin `Suppress("unused")` and Java `@SuppressWarnings("unused")`. In Java, Android Studio will help you add the Java annotation if you click Code -> Inspect Code then find a warning and click "Suppress". The Kotlin annotation will add a `/** @noinspection unused*/` comment but I think the annotation `Suppress("unused")` is better than the comment. However, the Android Studio "Suppress" tool is probably the easiest way to find the name of the warning that you want to suppress, since this should work for other types of warnings (not just unused code).

Just to be clear, are you asking me to change all the `@Suppress("UNUSED")` / `@SuppressWarnings("UNUSED")` to lowercase "unused"? Or are you just saying "that would clear the Android Studio warning but it's not necessary for this PR"?

asolovay:
> Just to be clear, are you asking me to change all the `@Suppress("UNUSED")` / `@SuppressWarnings("UNUSED")` to lowercase "unused"? Or are you just saying "that would clear the Android Studio warning but it's not necessary for this PR"?

Per our chat, I'll change all the suppress parameters to lower-case and re-push, stay tuned!

asolovay:
It looks like changing to "unused" fixed that warning! (I have one other unrelated warning, but it's about code that's on DAC so I'll want to do it in a separate CL where I can get the SME's approval.)

Let me know if you see any more issues.

PR #614, opened 2025-09-02
Add android_wear_tile_version_fallback

ithinkihaveacat:


ithinkihaveacat:
FYI @NedaTop.

wear/src/main/java/com/example/wear/snippets/tile/Tile.kt
```diff
@@ -200,3 +207,42 @@ class DynamicHeartRate : TileService() {
...
+class FeatureFallback : TileService() {
+    override fun onTileRequest(requestParams: RequestBuilders.TileRequest): ListenableFuture<Tile> {
+
+        // [START android_wear_tile_version_fallback]
+        val rendererVersion = requestParams.deviceConfiguration.rendererSchemaVersion
+
+        val arcElement =
+            if (
+                rendererVersion.major > 1 ||
+                (rendererVersion.major == 1 && rendererVersion.minor >= 500)
```
garanj, line 222: What's the significance of the 1 and 500 - can these be extracted to descriptive constants?

garanj (approved):

wear/src/main/java/com/example/wear/snippets/tile/Tile.kt
ithinkihaveacat: I don't think there's any way to usefully move them to constants. I moved and improved a comment to hopefully provide better context though, see 5ac68742c682baacdf32acf2e0352c35ad220cfa.

PR #620, opened 2025-09-11
Adding shadows code snippets

wardlauren:
Adding code snippets for Modifier.shadow and new shadows modifiers (dropShadow, innerShadow).


compose/snippets/src/main/java/com/example/compose/snippets/graphics/Shadows.kt
```diff
@@ -0,0 +1,609 @@
...
+                fontSize = 32.sp
+            )
+        }
+    }
+}
+// [END android_compose_graphics_simple_drop_shadow]
+
+// [START android_compose_graphics_simple_inner_shadow]
+@Composable
+@Preview(
```
riggaroo, line 118: You can move the @Preview to be before the [START] tags for all of these :)

PR #631, opened 2025-09-27
feat(views): Add basic notification snippet

askandroid:
Add a basic notification snippet and associated files to the views module.

Region-Tag: android_views_notifications_build_basic

Testing workflow for GitHub code migration.


views/build.gradle.kts
```diff
@@ -48,6 +48,8 @@ android {
 
 dependencies {
 
+    implementation("androidx.core:core:1.13.1")
```
cartland, line 51: We should put these imports in the version catalog, like the other imports.

PR #649, opened 2025-10-08
Add Gemini Developer API code snippets

cartland:
This commit introduces a collection of code snippets demonstrating the usage of the Gemini Developer API. It includes examples for model configuration, text-only and multimodal inputs (image, audio, video), multi-turn chat, and image generation/editing.

The necessary dependencies for Firebase AI, Guava, and Reactive Streams have been added to gradle/libs.versions.toml and misc/build.gradle.kts.

Intended documentation pages:
* https://developer.android.com/ai/gemini/developer-api
* https://developer.android.com/ai/imagen


gradle/libs.versions.toml
```diff
@@ -40,10 +40,13 @@ compose-latest = "1.9.3"
 composeUiTooling = "1.5.3"
 coreSplashscreen = "1.0.1"
 coroutines = "1.10.2"
+firebase-bom = "34.3.0"
```
lethargicpanda: Can we use `34.4.0`

misc/src/main/java/com/example/snippets/ai/GeminiDeveloperApiSnippets.kt
```diff
@@ -0,0 +1,204 @@
...
+    // Start by instantiating a GenerativeModel and specifying the model name:
+    val model = Firebase.ai(backend = GenerativeBackend.googleAI())
+        .generativeModel("gemini-2.5-flash")
+    // [END android_gemini_developer_api_gemini_25_flash_model]
+}
+
+object Gemini25FlashImagePreviewModelConfiguration {
+    // [START android_gemini_developer_api_gemini_25_flash_image_model]
+    val model = Firebase.ai(backend = GenerativeBackend.googleAI()).generativeModel(
+        modelName = "gemini-2.5-flash-image-preview",
```
lethargicpanda, line 43: Is there a way to use _dev site_ variables in our snippets?
I want us to try to move away from hardcoding the model names in our DAC code blocks.

gradle/libs.versions.toml
sanbeiji: Would it make sense to use devsite variables for these values?

misc/src/main/java/com/example/snippets/ai/GeminiDeveloperApiSnippets.kt
sanbeiji: Same question. :) Further to this point, we're sharing values with other teams. Can we just add those variables in here or will that cause questionable results?

gradle/libs.versions.toml
cartland: I updated to 34.4.0 in the code. We can talk about variables in the documentation, since GitHub cannot use those variables here.

misc/src/main/java/com/example/snippets/ai/GeminiDeveloperApiSnippets.kt
cartland: The documentation does not currently use variables. If you want to turn those into variables, we could make a separate standalone hard-coded snippet in the documentation that uses the documentation tool variables, but we cannot use the same variables in GitHub. We could then modify the GitHub code to use the same variable name, and define the variable somewhere outside of the snippet region tags.

cartland: @sanbeiji do you have any additional feedback on this? Since there aren't any equivalent documentation variables to consider right now, perhaps we can address that later if those variables are created?

PR #652, opened 2025-10-09
Add location snippet

kul3r4:
This adds a snippet for https://developer.android.com/training/wearables/apps/location-detection.md#kotlin

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/m3/navigation/Navigation.kt
+++ b/wear/src/main/java/com/example/wear/snippets/m3/navigation/Navigation.kt
@@ -74,7 +74,7 @@ fun MessageDetail(id: String) {
         contentPadding = padding
     ) { scaffoldPaddingValues ->
         // Screen content goes here
-        // [END android_wear_navigation]
+        // [START_EXCLUDE]
         TransformingLazyColumn(
             state = scrollState,
             contentPadding = scaffoldPaddingValues
@@ -87,6 +87,8 @@ fun MessageDetail(id: String) {
                 )
             }
         }
+        // [END_EXCLUDE]
+        // [END android_wear_navigation]
     }
 }
 


wear/src/main/java/com/example/wear/snippets/location/LocationActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        // [START_EXCLUDE]
+        setContent {
+            WearApp(hasGps = { hasGps() })
+        }
+        // [END_EXCLUDE]
+    }
+    fun hasGps(): Boolean =
```
garanj, line 47: How do hasSystemFeature(PackageManager.FEATURE_LOCATION_GPS) and LocationManagerCompat.isLocationEnabled(locationManager) interoperate and is the latter relevant here?

For example, if the watch has GPS, but Location is disabled in the settings, do these return true and false respectively, and does that matter in this use case?

Or does FEATURE_LOCATION_GPS already take into account whether location is enabled by the location setting?

kul3r4: Thanks for raising, it does NOT take into account whether location is enabled by the user setting. The current documentation doesn't mention it but I can add a new section and also mention this.

kul3r4: rereading https://developer.android.com/training/wearables/apps/location-detection#DetectGPS, this is for GPS only and there is another section for requesting permissions and in https://developer.android.com/develop/sensors-and-location/location/retrieve-current?#last-known (which is linked from the same page) it explains what happens when location is tuned off in settings. 

In summary: I think this snippet works as it is as this section is only focusing on gps, other sections of the same page cover the other cases.

PR #668, opened 2025-10-23
Add sample for WebView force dark mode

sushantsikka:


sushantsikka:
@cartland could you please check if these file locations make sense?

misc/src/main/java/com/example/snippets/jetpackwebkit/JetpackWebkitActivity.kt
```diff
@@ -0,0 +1,37 @@
+package com.example.webviewdarkmode
```
cartland: The package should match the directory structure after /java. In this case, it should be `com.example.snippets.jetpackwebkit`

misc/src/main/java/com/example/snippets/jetpackwebkit/JetpackWebkitActivity.kt
```diff
@@ -0,0 +1,37 @@
...
+        // This is necessary to keep navigation inside your WebView
+        webView.webViewClient = WebViewClient()
+
+        // Enable JavaScript
+        webView.settings.javaScriptEnabled = true
+
+        // To enable a modern feature, you pass that instance to a Jetpack Webkit helper.
+        if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)) {
+            WebSettingsCompat.setForceDark(webView.settings, WebSettingsCompat.FORCE_DARK_ON)
+        }
```
cartland, line 36: This snippet is supposed to match the snippet here:
https://developer.android.com/develop/ui/views/layout/webapps/jetpack-webkit-overview

```
// You still get your WebView instance the standard way.
val webView: WebView = findViewById(R.id.my_webview)

// To enable a modern feature, you pass that instance to a Jetpack Webkit helper.
if (WebViewFeature.isFeatureSupported(WebViewFeature.FORCE_DARK)) {
    WebSettingsCompat.setForceDark(webView.settings, WebSettingsCompat.FORCE_DARK_ON)
}
```

This snippet also adds additional code:

```
        // This is necessary to keep navigation inside your WebView
        webView.webViewClient = WebViewClient()

        // Enable JavaScript
        webView.settings.javaScriptEnabled = true
```

Is this additional code expected and required?

misc/src/main/java/com/example/snippets/jetpackwebkit/app/build.gradle.kts
cartland, line 1: We should not add a new module. We should edit the build file for the existing misc module.

https://github.com/android/snippets/blob/main/misc/build.gradle.kts

misc/src/main/java/com/example/snippets/jetpackwebkit/gradle/libs.versions.toml
cartland, line 1: We should not add a new Gradle version catalog. We should edit the existing version catalog.

https://github.com/android/snippets/blob/main/gradle/libs.versions.toml

misc/src/main/res/layout/jetpack_webkit_layout.xml
cartland, line 1: The name of this file appears to be a layout but the contents is an Android manifest file.

The code snippet uses `R.layout.activity_main` so you need the manifest file to match, `activity_main.xml`. 

https://developer.android.com/guide/topics/resources/layout-resource

Then the contents will need to be updated to be a layout. For the snippet to compile, it can be a very simple layout file.

misc/src/main/java/com/example/snippets/jetpackwebkit/JetpackWebkitActivity.kt
```diff
@@ -0,0 +1,37 @@
...
+class MainActivity : AppCompatActivity() {
+    @SuppressLint("SetJavaScriptEnabled")
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        // Sets the user interface from a layout resource.
+        setContentView(R.layout.activity_main)
+
+        // You still get your WebView instance the standard way.
+        // [START android_views_notifications_build_basic]
+        val webView: WebView = findViewById(R.id.my_webview)
```
cartland, line 25: The original snippet has the imports. To include the snippets in the region tag, it would be easiest to include them as comments here. Region tags around the real imports above would not be stable because our style formatting tools in Android Studio and Spotless will reorder imports and comments, which would break region tags if you try to use them up there.

```
// import android.webkit.WebView
// import androidx.webkit.WebSettingsCompat
// import androidx.webkit.WebViewFeature
```

misc/src/main/java/com/example/snippets/jetpackwebkit/JetpackWebkitActivity.kt
sushantsikka: It wouldn't harm to retain these lines of code, so prefer to retain.

misc/src/main/java/com/example/snippets/jetpackwebkit/JetpackWebkitActivity.kt
sushantsikka: Done.

misc/src/main/java/com/example/snippets/jetpackwebkit/JetpackWebkitActivity.kt
sushantsikka: Done.

PR #686, opened 2025-11-07
Migrate Gemini overview and Vertex AI Gemini API snippets

kkuan2011:
Code snippets are from these existing pages:
https://developer.android.com/ai/gemini
https://developer.android.com/ai/vertex-ai-firebase

### DIFF
--- a/misc/src/main/java/com/example/snippets/ai/VertexAiGeminiApi.kt
+++ b/misc/src/main/java/com/example/snippets/ai/VertexAiGeminiApi.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


misc/src/main/java/com/example/snippets/ai/VertexAiGeminiApi.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.coroutines.launch
+
+class VertexAiGeminiApi {
+
+    // [START android_snippets_vertex_ai_gemini_api_model]
+    val model = Firebase.ai(backend = GenerativeBackend.vertexAI())
+        .generativeModel("gemini-2.5-flash")
+    // [END android_snippets_vertex_ai_gemini_api_model]
+
+    fun generateText(scope: CoroutineScope) {
```
ksemenova, line 32: it would be better to made this a "suspend" function and remove "scope", but since the full code is not used in snippets we can ignore it.

ksemenova (approved):

misc/src/main/java/com/example/snippets/ai/VertexAiGeminiApi.kt
kkuan2011: ok noted for next time, thank you Kate!

PR #688, opened 2025-11-12
Add snippets for trigger based profiling section of ProfilingManager docs

edgararriagag:
Snippets for upcoming trigger based profiling section

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -14,7 +14,8 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
@@ -344,6 +345,36 @@
             </intent-filter>
             <!-- [END android_wear_complication_intent_filter] -->
         </activity>
+
+        <service
+            android:name=".snippets.m3.tile.StateTile"
+            android:label="@string/tile_label"
+            android:description="@string/tile_description"
+            android:icon="@mipmap/ic_launcher"
+            android:exported="true"
+            android:permission="com.google.android.wearable.permission.BIND_TILE_PROVIDER">
+            <intent-filter>
+                <action android:name="androidx.wear.tiles.action.BIND_TILE_PROVIDER" />
+            </intent-filter>
+
+            <meta-data android:name="androidx.wear.tiles.PREVIEW"
+                android:resource="@drawable/tile_preview" />
+        </service>
+
+        <!-- [START android_wear_datalayerlistener_intent_filter] -->
+        <service
+            android:name=".snippets.datalayer.DataLayerListenerService"
+            android:exported="true"
+            tools:ignore="ExportedService" >
+            <intent-filter>
+                <action android:name="com.google.android.gms.wearable.DATA_CHANGED" />
+                <data
+                    android:scheme="wear"
+                    android:host="*"
+                    android:path="/start-activity" />
+            </intent-filter>
+        </service>
+        <!-- [END android_wear_datalayerlistener_intent_filter] -->
 
     </application>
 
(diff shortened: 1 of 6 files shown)


misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -69,5 +77,39 @@ public void accept(ProfilingResult profilingResult) {
...
+          ProfilingTrigger.TRIGGER_TYPE_APP_FULLY_DRAWN);
+      triggerBuilder.setRateLimitingPeriodHours(1);
+      triggers.add(triggerBuilder.build());
+
+      Executor mainExecutor = Executors.newSingleThreadExecutor();
+      Consumer<ProfilingResult> resultCallback =
+          new Consumer<ProfilingResult>() {
+            @Override
+            public void accept(ProfilingResult profilingResult) {
+              // ...
```
AliceYuan: what code is supposed to be here?
Suggest adding it as an empty method name here instead of //...

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -69,5 +77,39 @@ public void accept(ProfilingResult profilingResult) {
...
+              if (profilingResult.getErrorCode() == ProfilingResult.ERROR_NONE) {
+                Log.d(TAG,
+                    "Received profiling result. file: " + profilingResult.getResultFilePath());
+              }
+            }
+          };
+      profilingManager.registerForAllProfilingResults(mainExecutor, resultCallback);
+      profilingManager.addProfilingTriggers(triggers);
+
+      Choreographer.getInstance().postFrameCallback((f) -> {
```
AliceYuan, line 115: Do we want to put this into the snippet? This seems more like a test functionality to trigger the profiling trigger

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -69,5 +77,39 @@ public void accept(ProfilingResult profilingResult) {
...
+      Executor mainExecutor = Executors.newSingleThreadExecutor();
+      Consumer<ProfilingResult> resultCallback =
+          new Consumer<ProfilingResult>() {
+            @Override
+            public void accept(ProfilingResult profilingResult) {
+              // ...
+              if (profilingResult.getErrorCode() == ProfilingResult.ERROR_NONE) {
+                Log.d(TAG,
+                    "Received profiling result. file: " + profilingResult.getResultFilePath());
+              }
```
AliceYuan, line 108: maybe log failure if there is an error?

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
edgararriagag: Good call, I think its fine to omit, modified to exclude it.

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
edgararriagag: Done, I added some basic handling of response. I also added a call to setupUploadJob() which is the usual expectation here, which for now I put as a stub method and in an upcoming CL I will populate it because its part of a different snippet, for this snippet, I'd prefer if we just show the function call.

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
edgararriagag: Done

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -69,5 +77,53 @@ public void accept(ProfilingResult profilingResult) {
...
+      Executor mainExecutor = Executors.newSingleThreadExecutor();
+      Consumer<ProfilingResult> resultCallback =
+          new Consumer<ProfilingResult>() {
+            @Override
+            public void accept(ProfilingResult profilingResult) {
+              if (profilingResult.getErrorCode() == ProfilingResult.ERROR_NONE) {
+                Log.d(
+                    "ProfileTest",
+                    "Received profiling result file=" + profilingResult.getResultFilePath());
+                setupUploadJob(profilingResult.getResultFilePath());
```
AliceYuan: suggest renaming to `setupProfileUploadWorker`? to suggest to developer to use workmanager

AliceYuan (approved):
lgtm!

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
edgararriagag: sgtm, done

PR #699, opened 2025-11-20
Add snippets for /training/wearables/tiles/versioning

ithinkihaveacat:


wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
```diff
@@ -0,0 +1,112 @@
...
+import android.content.Context
+import androidx.wear.protolayout.DeviceParametersBuilders
+import androidx.wear.protolayout.DimensionBuilders.expand
+import androidx.wear.protolayout.LayoutElementBuilders.Column
+import androidx.wear.protolayout.ModifiersBuilders
+import androidx.wear.protolayout.TypeBuilders
+import androidx.wear.protolayout.material.Text
+import androidx.wear.protolayout.material.Typography
+import androidx.wear.protolayout.material.layouts.PrimaryLayout
+import androidx.wear.protolayout.material3.MaterialScope
```
kul3r4, line 28: flagging that this import is repeated twice, or is that on purpose?

wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
```diff
@@ -0,0 +1,112 @@
...
+        context: Context,
+        deviceConfiguration: DeviceParametersBuilders.DeviceParameters,
+    ) =
+        materialScope(context, deviceConfiguration) {
+            primaryLayout(mainSlot = { text("Hello, World!".layoutString) }) // HELPME
+        }
+    // [END android_wear_tile_version_layoutm3]
+
+    // [START android_wear_tile_version_modifierm3]
+    // The equivalent Compose-like modifier is much simpler
```
kul3r4: Can you explain why we need this comment? is that useful for learning this code?

kul3r4 (changes requested):

wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
ithinkihaveacat: It's to contrast with the M2.5 version. This is the context, does this make more sense?https://developer.android.com/training/wearables/tiles/versioning#modifiers

wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
ithinkihaveacat: They look the same but one is "MaterialScope" and the other is "materialScope"! (Both are needed.)

wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
kul3r4: ok, looking at the DAC page helped.
To me the comment you have in the 2.5 is useful as it explains what this does"// A Builder-style modifier to set the opacity of an element to 0.5" while in the M3 snippet you use the comment for highlighting the simplicity. If we ever remove the 2.5 snippet, this comment by itself in the snippet for M3 is not self explanatory, that was my thinking.

wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
kul3r4: thank!

wear/src/main/java/com/example/wear/snippets/tile/Versioning.kt
ithinkihaveacat: Oh I see. Ok, I modified the comment to make it more self-contained. (It will still will read a bit funny without the M2.5 snippet for contrast but it's much better.)

kul3r4: thanks

kul3r4 (approved):

PR #706, opened 2025-11-25
ProfilingManager docs V2: snippets for ANR section

edgararriagag:


misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -125,5 +128,85 @@ public void setupProfileUploadWorker(String resultFilePath) {
...
+      Consumer<ProfilingResult> resultCallback =
+          profilingResult -> {
+            // Handle uploading trace to your back-end
+          };
+      profilingManager.registerForAllProfilingResults(mainExecutor, resultCallback);
+      profilingManager.addProfilingTriggers(triggers);
+    }
+    // [END android_profiling_manager_anr_case_study_java_snippet_1]
+
+    // [START android_profiling_manager_anr_case_study_java_snippet_2]
```
AliceYuan, line 148: Not sure which section this is intended for, this doesn't seem like it would be related to ANR triggers? Looks like a manual trigger

edgararriagag: Ah... yeah, I think it is better to review this starting from the upcoming docs themselves as they describe the case study flow. I shared the doc with you offline for this, if this comment still stands after reviewing the docs, feel free to reopen.

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -125,5 +128,85 @@ public void setupProfileUploadWorker(String resultFilePath) {
...
+    }
+
+    public void onClickSubmit() {
+      prepareNetworkRequest();
+
+      boolean networkRequestSuccess = false;
+      int maxAttempts = 10;
+      while (!networkRequestSuccess && maxAttempts > 0) {
+        networkRequestSuccess = performNetworkRequest(NETWORK_TIMEOUT_MILLISECS);
+        maxAttempts--;
```
AliceYuan, line 164: Not sure if this logic `while (!networkRequestSuccess && maxAttempts > 0)` is necessary to share, I would just call `performNetworkRequest(NETWORK_TIMEOUT_MILLISECS)` without the reattempt logic. The reason being is that the reattempt logic would normally be in an asynchronous thread, so you wouldn't return attempt and retry on the main thread.

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
```diff
@@ -125,5 +128,85 @@ public void setupProfileUploadWorker(String resultFilePath) {
...
+    public void setupButtonCallback() {
+      findViewById(R.id.submit).setOnClickListener(submitButtonView -> {
+        Trace.beginSection("MyApp:SubmitButton");
+        onClickSubmit();
+        Trace.endSection();
+      });
+    }
+
+    public void onClickSubmit() {
+      prepareNetworkRequest();
```
AliceYuan, line 158: what would be in the `prepareNetworkRequest` method? It's not clear to me so maybe just remove for simplicity?

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
edgararriagag: Ah... this is actually intentional, because in the case study, we show this as part of the wrong use, we are really showing a bad pattern and how we can sport it with traces, so the reattempt logic is part of the cause and we explain why it causes the problem

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerJavaSnippets.java
edgararriagag: in the real world, it is expected to have any kind of data collection to prepare to be sent as part of the network request. It is part of the story telling required to build the idea behind whats going on, in this specific case, we just have it as a methood that consumes CPU but we don't specify the contents because its outside of the scope of the actual case study, we just put it so that in the trace we had certain computations consuming CPU time, it is part of the story telling required to showcase the study. Unfortunately, removing it would modify the trace data that was built to generate the case study, we could at most rename it. Though, I think the current name is still reasonable for the case study. I think we should leave as-is.

yrezgui (approved):

PR #716, opened 2025-12-01
Add snippet for b/454723857

amyZepp:


xr/src/main/java/com/example/xr/compose/Subspace.kt
```diff
@@ -64,3 +66,14 @@ private fun PrimaryPane() {
 
 @Composable
 private fun SecondaryPane() {}
+
+@Composable
+private fun RecommendedContentBoxInFullSpace() {
+
+    // [START androidxr_compose_RecommendedContentBoxInFullSpace]
+
+    val session = LocalSession.current
```
devbridie, line 73: Delete blank lines between START and END

amyZepp (commented):
updated

devbridie (approved):

PR #718, opened 2025-12-02
Update CODEOWNERS with new ownership rules

yrezgui:
This pull request makes a small update to the `CODEOWNERS` file to clarify team ownership for several directories and the root of the repository. The changes assign specific teams as owners for the root and for feature-specific directories.

### DIFF
--- a/xr/src/main/java/com/example/xr/compose/Subspace.kt
+++ b/xr/src/main/java/com/example/xr/compose/Subspace.kt
@@ -23,6 +23,7 @@ import androidx.activity.enableEdgeToEdge
 import androidx.compose.foundation.layout.Row
 import androidx.compose.runtime.Composable
 import androidx.xr.compose.platform.LocalSession
+import androidx.xr.compose.spatial.PlanarEmbeddedSubspace
 import androidx.xr.compose.spatial.Subspace
 import androidx.xr.compose.subspace.SpatialPanel
 import androidx.xr.scenecore.scene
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+*               @yrezgui @kkuan2011 @riggaroo @jakeroseman
+/compose/       @riggaroo @simona-anomis @wardlauren
+/build/         @amylgu @alexmikeck
+/wear/          @kul3r4 @shumelchyk @khufnagle
+/xr/            @vad710
```
kkuan2011: amyZepp

CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+*               @yrezgui @kkuan2011 @riggaroo @jakeroseman
```
kkuan2011: actually discussed with Rebecca, and we can remove her from the root

riggaroo: Suggest that we don't do this - I was getting a lot of messages from teams asking for a review approval on a snippet that had no relation to my area, this will overwhelm the four of us if we enforce approval from code owner outside of the four well defined groups here. 

Separately,  @telpirion was suggesting linking up these to MDB groups instead of individuals which i agree with as people move around teams and then this will get outdated.

yrezgui: We need fallback just in case but will definitely add owners per folder. I'm removing you from root and keeping others

PR #733, opened 2025-12-08
Fix code to show imports in Wear migration guidance

kul3r4:
Readd imports that were lost in previous PRs to show the imports that are useful in the Wear M3 migration guidance.

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/m3/theme/Color.kt
+++ b/wear/src/main/java/com/example/wear/snippets/m3/theme/Color.kt
@@ -18,9 +18,20 @@ package com.example.wear.snippets.m3.theme
 
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.platform.LocalContext
+// [START android_wear_color]
 import androidx.wear.compose.material3.ColorScheme
+// [START_EXCLUDE]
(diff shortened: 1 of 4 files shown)


wear/src/main/java/com/example/wear/snippets/m3/theme/Color.kt
```diff
@@ -18,10 +18,21 @@ package com.example.wear.snippets.m3.theme
 
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.platform.LocalContext
+// [START android_wear_color]
 import androidx.wear.compose.material3.ColorScheme
```
ithinkihaveacat, line 22: Being able to see the exact import is a good benefit of snippets, but I wonder if they can just be left in the GH source, rather than being visible on DAC. Among other things, you can only do this for the first snippet in the file (because of the way the EXCLUDE mechanism works).

kul3r4: agree this approach has limitations and I have been trying to find other ways without much success. This case for showing imports is only for the migration guidance where we wanted to show the imports for clarity but unfortunately imports cannot be duplicated in the same file, so the only alternative is to create a new file for each one

PR #735, opened 2025-12-09
do-not-review first commit play billing snippets 09122025

kumann-dev:
first commit play billing snippets 09122025

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/m3/theme/Color.kt
+++ b/wear/src/main/java/com/example/wear/snippets/m3/theme/Color.kt
@@ -18,9 +18,20 @@ package com.example.wear.snippets.m3.theme
 
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.platform.LocalContext
+// [START android_wear_color]
 import androidx.wear.compose.material3.ColorScheme
+// [START_EXCLUDE]
 import androidx.wear.compose.material3.MaterialTheme
 import androidx.wear.compose.material3.dynamicColorScheme
+
+
+fun color() {
+// [END_EXCLUDE]
+    val appColorScheme: ColorScheme = ColorScheme(
+        // M3 ColorScheme parameters
+    )
+    // [END android_wear_color]
+}
 
 // [START android_wear_dynamic_theme]
 @Composable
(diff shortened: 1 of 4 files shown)


kumann-dev (commented):
First commit of play billing snippets that will be used in the https://developer.android.com/google/play/billing DAC pages

play-billing-library/app/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+}
+
+
+
+dependencies {
+
+    // OSS Plugin
+    implementation 'com.google.android.gms:play-services-oss-licenses:17.1.0'
```
yrezgui, line 57: The dependencies version must be defined in the Gradle version catalog instead of being hard-coded here to avoid different versions across our modules

play-billing-library/build.gradle
```diff
@@ -0,0 +1,45 @@
+/* Copyright 2022 Google LLC
```
yrezgui, line 1: This whole folder is an Android Studio project. You should create a module within the existing project at the root level

play-billing-library/app/src/main/java/com/google/play/billing/samples/managedcatalogue/billing/BillingServiceClient.java
```diff
@@ -0,0 +1,160 @@
...
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+/**
+ * Manages interactions with the Google Play Billing Library for handling pre-orders.
+ *
+ * <p>This class encapsulates the setup of the {@link BillingClient}, manages the connection
+ * lifecycle (including retries), queries product details, and notifies a listener.
+ */
+public class BillingServiceClient {
```
yrezgui, line 41: Are the code snippets expected to be migrated to Kotlin anytime soon?

yrezgui (changes requested):

play-billing-library/build.gradle
kumann-dev: Can you please help me understand this comment? Do we already have an Android app in which we can add our play billing snippets OR do we have to create a new Android app specifically for Play Billing?

play-billing-library/app/src/main/java/com/google/play/billing/samples/managedcatalogue/billing/BillingServiceClient.java
kumann-dev: There is no migration required. We need have both Kotlin and Java snippets. For now, we want to get all the Java snippets in Github. Later we will put all the Kotlin snippets.

play-billing-library/build.gradle
yrezgui: You need to open the root folder as a project in Android Studio and create a module in that existing project

PR #741, opened 2025-12-12
Add snippets for new Compose State DAC guides

andrewbailey:
Snippets for both "State Lifespans" and "State Callbacks" DAC guides, to ship in [cl/842788892](http://cl/842788892).

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/state/RememberAndRetainObserverSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/state/RememberAndRetainObserverSnippets.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/state/RememberAndRetainObserverSnippets.kt
```diff
@@ -0,0 +1,211 @@
...
+        override fun onForgotten() {}
+        override fun onRemembered() {}
+    }
+
+    // [START android_compose_state_observers_initialize_in_remember]
+    class MyComposeObject : RememberObserver {
+        private val job = Job()
+        private val coroutineScope = CoroutineScope(Dispatchers.Main + job)
+
+        init {
```
riggaroo, line 39: Discussed the "not recommended" pattern a bit with @wardlauren, I think lets not commit it to the repository, and rather hardcode the not recommended pattern snippet, as we dont want people to copy/ paste this.

andrewbailey: This snippet has a mix of both recommended and not recommended. Did you want to split this up or are you just referring to the "fully not recommended" snippets?

If we only remove the fully not recommended snippets, that takes out these from the PR:
 - `android_compose_state_observers_public_implementation`
 - `android_compose_state_observers_only_remember_once`
 - `android_compose_state_observers_reremember_params`

andrewbailey: Discussed offline with @wardlauren — let's keep this mixed sample as-is. Splitting it into a recommended and not-recommended snippet would result in a lot of code duplication and would look very verbose.

PR #745, opened 2025-12-17
Create authentication snippets

JohnZoellerG:
Create snippets for Wear Auth DAC: https://developer.android.com/training/wearables/apps/auth-wear

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/auth/CredentialManager.kt
+++ b/wear/src/main/java/com/example/wear/snippets/auth/CredentialManager.kt
@@ -59,7 +59,7 @@ class CredentialManagerAuthenticator(
         return false
     }
 
-/**signInRequest
+    /**
      * Creates a [GetCredentialRequest] with standard Wear Credential types.
      *
      * @return A configured [GetCredentialRequest] ready to be used with [CredentialManager.getCredential].
@@ -75,7 +75,7 @@ class CredentialManagerAuthenticator(
         )
     }
 
-/**
+    /**
      * Routes the credential received from `getCredential` to the appropriate authentication
      * type handler on the [AuthenticationServer].
      *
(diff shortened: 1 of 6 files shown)


wear/src/main/java/com/example/wear/snippets/auth/WristDetectionUtility.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import android.app.KeyguardManager
+import android.content.Context
+import androidx.core.content.getSystemService
+
+// This setting differs across OEMs, so you may want to check several in your detection function
+// This one works for the pixel watch.
+private const val PIXEL_WRIST_AUTOLOCK_SETTING_STATE = "wrist_detection_auto_locking_enabled"
+
+// [START android_wear_auth_wrist_detection]
```
kul3r4, line 25: is there an END tag for this?

JohnZoellerG: Yes, its just hard to see as the first indented line of the function.  It may seem strange to only provide the function signature, but this is how the auth doc looks today and I'd rather abstract away the details than having the reader take a detour into keyguard manager.

wdyt?

kul3r4: ah sorry totally missed it!

wear/src/main/java/com/example/wear/snippets/auth/CredentialManager.kt
```diff
@@ -0,0 +1,135 @@
...
+                credentialManager.getCredential(activity, createGetCredentialRequest())
+            return authenticate(getCredentialResponse.credential)
+        } catch (_: GetCredentialCancellationException) {
+            navigateToSecondaryAuthentication()
+        }
+        // [END android_wear_credential_manager_secondary_fallback]
+        return false
+    }
+
+/**signInRequest
```
kkuan2011: do you need "signInRequest" here?

JohnZoellerG: nope good catch!

PR #746, opened 2025-12-17
ProfilingManager Docs V2: Add trace upload job snippets

edgararriagag:


misc/build.gradle.kts
```diff
@@ -36,8 +36,8 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_17
-        targetCompatibility = JavaVersion.VERSION_17
+        sourceCompatibility = JavaVersion.VERSION_21
```
AliceYuan: If this isn't a necessity for the PR I would remove this change, we don't know for sure that there's no snippets in the codebase that have dependencies of lower java compat versions.

misc/build.gradle.kts
```diff
@@ -21,7 +21,7 @@ android {
     }
 
     kotlin {
-        jvmToolchain(17)
+        jvmToolchain(21)
```
AliceYuan: same here about undo upgrade

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerKotlinSnippets.kt
```diff
@@ -137,4 +146,77 @@ class ProfilingManagerKotlinSnippets {
...
+        class TraceUploadWorker(context: Context, workerParams: WorkerParameters) : Worker(context, workerParams) {
+            override fun doWork(): Result {
+                // Perform your uploading work here
+                Log.d("ProfileTest", "Uploading trace: " + inputData.getString("PROFILE_PATH"))
+
+                return Result.success()
+            }
+        }
+
+        fun setupProfileUploadWorker(profileFilepath: String?) {
```
AliceYuan, line 152: the sample seems unnecessarily long, can we just share the worker code instead of sharing the record system trace section as well which we've already provided in the previous traces?

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerKotlinSnippets.kt
```diff
@@ -137,4 +146,77 @@ class ProfilingManagerKotlinSnippets {
...
+
+            val inputDataBuilder = Data.Builder()
+            inputDataBuilder.putString("PROFILE_PATH", profileFilepath)
+            workRequestBuilder.setInputData(inputDataBuilder.build())
+
+            workMgr.enqueue(workRequestBuilder.build())
+        }
+
+        @RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+        fun sampleRecordSystemTraceWithJob() {
```
AliceYuan: We already have `sampleRecordSystemTrace` function, can we just update it to call `setupProfileUploadWorker` and then show the upload worker as a separate code snippet? That way we don't have excessive code in the code sample and they can still click into the github repo to see the full snippet.

misc/build.gradle.kts
edgararriagag: Agreed. Done

misc/build.gradle.kts
edgararriagag: Done

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerKotlinSnippets.kt
edgararriagag: Great call, done

misc/src/main/java/com/example/snippets/profiling/ProfilingManagerKotlinSnippets.kt
edgararriagag: Agreed. Done

PR #747, opened 2025-12-18
Adds watch face snippets

garanj:
Creates Watch Face Format snippets for use in the watch face documentation on developer.android.com

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -29,6 +29,7 @@ androidx-test-junit = "1.3.0"
 androidx-window = "1.5.1"
 androidx-window-core = "1.5.1"
 androidx-window-java = "1.5.1"
+androidx-work-runtime = "2.11.0"
 androidx-xr-arcore = "1.0.0-alpha09"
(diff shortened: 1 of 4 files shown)


watchface/src/main/res/raw/watchface.xml
```diff
@@ -0,0 +1,28 @@
...
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+<WatchFace width="450" height="450">
+  <Scene>
+      <!-- Content to be rendered -->
+      <!-- [START_EXCLUDE] -->
```
kul3r4: qq why do you need exclude here if there is no snippet?

garanj: Poor copy and paste skills! Thanks!

kul3r4: thanks!

PR #751, opened 2025-12-22
Add AI glasses projected snippets

trambui09:
Include:

- Create glasses activity
- Launching a glasses activity in a phone main activity
- Gate launch of glasses activity via `ProjectedContext.isProjectedDeviceConnected()` method


gradle/libs.versions.toml
```diff
@@ -98,6 +98,8 @@ wearToolingPreview = "1.0.0"
 webkit = "1.14.0"
 wearPhoneInteractions = "1.1.0"
 wearRemoteInteractions = "1.1.0"
+glimmer = "1.0.0-alpha03"
```
kkuan2011: could we prefix these new variables with xr so we know they're for XR? such as xrGlimmer?

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.xr.glimmer.Button
+import androidx.xr.glimmer.Card
+import androidx.xr.glimmer.GlimmerTheme
+import androidx.xr.glimmer.Text
+import androidx.xr.glimmer.surface
+
+// [START androidxr_projected_ai_glasses_activity]
+class GlassesMainActivity : ComponentActivity() {
```
kkuan2011, line 46: Since you already have the code for an activity, do you want to add these activities to the manifest file for the xr module? If not, that's okay too, not a blocking comment.

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+            GlimmerTheme {
+                HomeScreen(onClose = {
+                    finish()
+                })
+            }
+        }
+    }
+
+    override fun onStart() {
+        super.onStart()
```
kkuan2011: Instead of overriding the activity lifecycle methods directly, I believe it's better to use a LifecycleObserver (https://developer.android.com/topic/libraries/architecture/lifecycle). Not a blocking comment, but could you look into this? That would make it more Compose/architecture best practice friendly.

xr/src/main/java/com/example/xr/projected/PhoneMainActivity.kt
```diff
@@ -0,0 +1,111 @@
...
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.xr.projected.ProjectedContext
+import androidx.xr.projected.experimental.ExperimentalProjectedApi
+
+class MainActivity : ComponentActivity() {
```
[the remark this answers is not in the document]
trambui09: Done! Thanks Gemini!

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+
+}
+// [END androidxr_projected_ai_glasses_activity]
+
+// [START androidxr_projected_ai_glasses_activity_homescreen]
+@Composable
+fun HomeScreen(modifier: Modifier = Modifier, onClose: () -> Unit) {
+    Box(
+        modifier = modifier
+            .surface(focusable = false).fillMaxSize(),
```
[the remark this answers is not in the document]
trambui09: Done!

gradle/libs.versions.toml
trambui09: Done!

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
trambui09: Ah, yes. This is important! And will be part of the snippets too. Added

trambui09: Is there a way to add the region tags in the manifest file?

Would it be?

```<!-- [START androidxr_projected_ai_glasses_activity_manifest]

[END androidxr_projected_ai_glasses_activity_manifest] -->```

File: gradle/libs.versions.toml [hunk 6]
```diff
@@ -98,6 +98,8 @@ wearToolingPreview = "1.0.0"
 webkit = "1.14.0"
 wearPhoneInteractions = "1.1.0"
 wearRemoteInteractions = "1.1.0"
+xrGlimmer = "1.0.0-alpha03"
+projected = "1.0.0-alpha03"
```
kkuan2011: nit: also change to xrProjected

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
kkuan2011: Ah per our snippets criteria, you can actually omit the region tags from the manifest and leave that hardcoded in the documentation pages. It would get more complicated as more docs need to reference parts of the manifest file and there's no way to have overlapping region tags.

gradle/libs.versions.toml
trambui09: ah yeah, I had it changed and forgot to push it up! done!

xr/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,23 @@
...
+            android:requiredDisplayCategory="@string/display_category_xr_projected">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+            </intent-filter>
+        </activity>
+        <activity
+            android:name="com.example.xr.projected.PhoneMainActivity"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
```
kkuan2011, line 35: nit: can you have two activities with the Main intent filter?

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,128 @@
...
+import androidx.xr.projected.experimental.ExperimentalProjectedApi
+import kotlinx.coroutines.launch
+
+// [START androidxr_projected_ai_glasses_activity]
+@OptIn(ExperimentalProjectedApi::class)
+class GlassesMainActivity : ComponentActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val viewModel: GlassesViewModel by viewModels()
```
dturner: It looks like you don't need to retain the UI state through config changes because it is recreated each time `onCreate` is called. In which case, you're better off declaring your UI state variables as `MutableState`s. For example: 

```
var isVisualSupported by mutableStateOf(false)
var areVisualsOn by mutableStateOf(false)
```

That way you don't need a ViewModel.

xr/src/main/java/com/example/xr/projected/PhoneMainActivity.kt
```diff
@@ -0,0 +1,111 @@
...
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.Center
+        ) {
+            Text(
+                text = "Hello AI Glasses",
+                style = MaterialTheme.typography.titleLarge
+            )
+            Spacer(modifier = Modifier.height(32.dp))
+            val scope = rememberCoroutineScope()
+            val isGlassesConnected by ProjectedContext.isProjectedDeviceConnected(
```
dturner: nit: "glasses" is plural so this should be `areGlassesConnected`

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,128 @@
...
+            val observer = GlassesLifecycleObserver(
+                context = this@GlassesMainActivity,
+                controller = displayController,
+                onVisualsChanged = viewModel::updateVisuals
+            )
+            lifecycle.addObserver(observer)
+
+
+            // Cleanup observer to close the display controller
+            lifecycle.addObserver(object : DefaultLifecycleObserver {
```
dturner: If `lifecycleScope` is cancelled before the cleanup observer is added, `onDestroy` won't be called. A safer approach would be to hoist `projectedDisplayController` outside of `onCreate` and make it nullable. Then inside `onCreate` but before calling `lifecycleScope.launch` add the observer: 

```
lifecycle.addObserver(object : DefaultLifecycleObserver {
  override fun onDestroy(owner: LifecycleOwner) {
    displayController?.close()
      displayController = null
    }
  }
)
```

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val viewModel: GlassesViewModel by viewModels()
+
+        lifecycleScope.launch {
+            // [START androidxr_projected_device_capabilities_check]
+            // Check device capabilities
+            val projectedDeviceController = ProjectedDeviceController.create(this@GlassesMainActivity)
+            val isVisualSupported = projectedDeviceController.capabilities.contains(CAPABILITY_VISUAL_UI)
```
dturner: nit: `isVisualUiSupported`

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    }
+}
+// [END androidxr_projected_ai_glasses_activity]
+
+// [START androidxr_projected_ai_glasses_activity_homescreen]
+@Composable
+fun HomeScreen(
+    visualsOn: Boolean,
```
dturner: nit: `areVisualsOn`

trambui09: Done!

xr/src/main/java/com/example/xr/projected/PhoneMainActivity.kt
trambui09: done!

PR #761, opened 2026-01-06
feat: adds a SharedPreferencesMigration example (preferences)

telpirion:
This PR adds the `android_datastore_preferences_migrate` region tag. Internal issue 446749253 tracks.

**Note**: I will add a corresponding snippet for proto & JSON preferences as I have time. Thanks!

### DIFF
--- a/wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnService.kt
+++ b/wear/src/main/java/com/example/wear/snippets/alwayson/AlwaysOnService.kt
@@ -16,12 +16,15 @@
 
 package com.example.wear.snippets.alwayson
 
+import android.Manifest
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.PendingIntent
+import android.content.Context
 import android.content.Intent
 import android.os.SystemClock
 import android.util.Log
+import androidx.annotation.RequiresPermission
 import androidx.core.app.NotificationCompat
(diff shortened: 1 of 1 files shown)


datastore/src/main/java/com/example/datastore/snippets/preferences/PreferencesDataStore.kt
```diff
@@ -30,6 +30,20 @@ import kotlinx.coroutines.flow.map
 val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = "settings")
 // [END android_datastore_preferences_create]
 
+/*
+// [START android_datastore_preferences_migrate]
```
kkuan2011: Thanks Eric! Should we move these migration snippets into a new file called PreferencesMigration or something? So that it doesn't need to be in a block comment?

telpirion: That's a good question and I was debating doing that. My concern is that the declaration will duplicate the declaration on line 30.

What do you think?

telpirion: You know what, I'm just going to take your suggestion and put this in a new file. That way we can show the import statement for `SharedPreferencesMigration` along with the code snippet.

Thanks Kat!

PR #763, opened 2026-01-07
Update Autofill Snippets to include both state value based TF

MagicalMeghan:
Updating from having two files to merging into one

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,3 +5,4 @@
 /wear/          @android/devrel-wear
 /wearcompanion/ @android/devrel-wear
 /xr/            @android/devrel-xr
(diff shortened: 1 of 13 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/text/AutofillSnippets.kt
```diff
@@ -40,6 +40,18 @@ import com.example.compose.snippets.touchinput.Button
...
 @Composable
 fun AddAutofill() {
+    val textFieldValue = remember {
+        mutableStateOf(TextFieldValue(""))
+    }
+
+    // [START android_compose_autofill_1_value]
+    TextField(
+        value = textFieldValue.value,
+        onValueChange = {textFieldValue.value = it},
```
riggaroo: Can you run spotless here?

PR #766, opened 2026-01-13
Adds WFP snippets

garanj:
Adds Snippets for Watch Face Push, for the app aspects (as opposed to the validation aspects).

This required changing the `minSdk` of the `wear` module to 33 as this is the minimum supported by Watch Face Push AndroidX library.

### DIFF
--- a/datastore/src/main/java/com/example/datastore/snippets/sharedprefsmigration/PreferencesDataStore.kt
+++ b/datastore/src/main/java/com/example/datastore/snippets/sharedprefsmigration/PreferencesDataStore.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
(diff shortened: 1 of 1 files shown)


wear/build.gradle.kts
```diff
@@ -13,7 +13,7 @@ android {
 
     defaultConfig {
         applicationId = "com.example.wear"
-        minSdk = 26
+        minSdk = 33
```
yrezgui, line 16: Why are we dropping the `minSdk` for Wear?

yrezgui (changes requested):

wear/build.gradle.kts
garanj: Watch Face Push library requires minSdk 33

yrezgui (approved):
Let's move the discussion on `minSdk` to a later PR

PR #769, opened 2026-01-15
Adds TV snippets

pflammertsma:


settings.gradle.kts
```diff
@@ -45,3 +45,4 @@ include(
     ":kmp:androidApp",
     ":kmp:shared"
 )
+include(":tv")
```
yrezgui: Add the module to the above list

tv/src/test/java/com/example/tv/ExampleUnitTest.kt
```diff
@@ -0,0 +1,2 @@
+package com.example.tv
```
yrezgui: Remove unused file

PR #774, opened 2026-01-22
Add missing snippet and add more code for formatting

kul3r4:
Follow up from #772:
- adding one missing snippet for manifest
- extending the snippet for datalayer otherwise format is odd, plus adds more informations
- adds snippets for https://developer.android.com/training/wearables/apps/standalone-apps.html


wear/src/main/AndroidManifest.xml
```diff
@@ -31,9 +31,9 @@
     <!-- Required to be able to call the setWatchFaceAsActive() method. -->
     <uses-permission android:name="com.google.wear.permission.SET_PUSHED_WATCH_FACE_AS_ACTIVE" />
     <!-- [END android_wear_wfp_active_permission] -->
-
+    <!-- [START android_wear_watch-feature] -->
```
garanj: Is there where the GitHub bot says there is a violation? Perhaps the use of "-" instead of "_"?

kul3r4: oh thank you good catch!!!

PR #777, opened 2026-01-22
Refactor DataLayer method to use coroutines rather than callback.

kul3r4:
Refactor DataLayer method to use coroutines rather than callback.


kkuan2011 (approved):
LGTM from snippets repo perspective

PR #808, opened 2026-02-24
Add a function to create a nonce for Sign in with Google.kt

vinishavathwani:
Add a snippet that demonstrates how to create a secure nonce. Also make minor other updates to the existing snippets based on the [draft doc](https://docs.google.com/document/d/1h5E6tyj5SQierwlaUjBuvWSKEsLhxmSo7BZ4ZtxpBHs/edit?tab=t.lv1y7e4mkj7z).

### DIFF
--- a/identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/SignInWithGoogleFunctions.kt
+++ b/identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/SignInWithGoogleFunctions.kt
@@ -55,7 +55,7 @@ class SignInWithGoogleFunctions(
         return googleIdOption
     }
 
-    private val googleIdOption = createGoogleIdOption("")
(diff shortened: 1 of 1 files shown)


identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/SignInWithGoogleFunctions.kt
kkuan2011: should we remove this parameter then if it's not used?

cy245: +1

vinishavathwani: Good point! Done.

PR #821, opened 2026-02-27
add in projected permissions and hardware access snippets

trambui09:
Projected Snippets:

- Updated GlassesMainActivity.kt to include hardware permission handling using the new ProjectedPermissionsResultContract.

Created ProjectedHardware.kt covering:

- Obtaining ProjectedContext for both glasses and host (phone) devices.
- High-fidelity image capture using CameraX with Camera2 Interop for FPS and resolution control.

### DIFF
--- a/xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
+++ b/xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
@@ -54,16 +54,18 @@ class GlassesMainActivity : ComponentActivity() {
     private var displayController: ProjectedDisplayController? = null
     private var isVisualUiSupported by mutableStateOf(false)
     private var areVisualsOn by mutableStateOf(true)
+    private var isPermissionDenied by mutableStateOf(false)
 
     // [START androidxr_projected_permissions_launcher]
     // Register the permissions launcher using the ProjectedPermissionsResultContract.
     private val requestPermissionLauncher: ActivityResultLauncher<List<ProjectedPermissionsRequestParams>> =
         registerForActivityResult(ProjectedPermissionsResultContract()) { results ->
             if (results[Manifest.permission.CAMERA] == true) {
-                // Permission granted, initialize the session/features.
+                isPermissionDenied = false
                 initializeGlassesFeatures()
             } else {
                 // Handle permission denial.
+                isPermissionDenied = true
             }
         }
     // [END androidxr_projected_permissions_launcher]
@@ -91,6 +93,8 @@ class GlassesMainActivity : ComponentActivity() {
                 HomeScreen(
                     areVisualsOn = areVisualsOn,
                     isVisualUiSupported = isVisualUiSupported,
+                    isPermissionDenied = isPermissionDenied,
+                    onRetryPermission = { requestHardwarePermissions() },
                     onClose = { finish() }
                 )
             }
@@ -140,6 +144,8 @@ class GlassesMainActivity : ComponentActivity() {
 fun HomeScreen(
     areVisualsOn: Boolean,
     isVisualUiSupported: Boolean,
+    isPermissionDenied: Boolean,
+    onRetryPermission: () -> Unit,
     onClose: () -> Unit,
     modifier: Modifier = Modifier
 ) {
@@ -149,7 +155,15 @@ fun HomeScreen(
             .fillMaxSize(),
         contentAlignment = Alignment.Center
     ) {
-        if (isVisualUiSupported) {
+        if (isPermissionDenied) {
+            Card(
+                title = { Text("Permission Required") },
+                action = { Button(onClick = onClose) { Text("Exit") } }
+            ) {
+                Text("Camera access is needed to use AI glasses features.")
+                Button(onClick = onRetryPermission) { Text("Retry") }
+            }
+        } else if (isVisualUiSupported) {
             Card(
                 title = { Text("Android XR") },
                 action = {


xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -49,6 +55,19 @@ class GlassesMainActivity : ComponentActivity() {
     private var isVisualUiSupported by mutableStateOf(false)
     private var areVisualsOn by mutableStateOf(true)
 
+    // [START androidxr_projected_permissions_launcher]
+    // Register the permissions launcher using the ProjectedPermissionsResultContract.
+    private val requestPermissionLauncher: ActivityResultLauncher<List<ProjectedPermissionsRequestParams>> =
+        registerForActivityResult(ProjectedPermissionsResultContract()) { results ->
```
kkuan2011, line 62: Should we use rememberLauncherForActivityResult() instead?
https://developer.android.com/develop/ui/compose/libraries#requesting-runtime-permissions

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
```diff
@@ -0,0 +1,141 @@
...
+ */
+// [START androidxr_projected_context_get_projected]
+@OptIn(ExperimentalProjectedApi::class)
+private fun getGlassesContext(context: Context): Context? {
+    return try {
+        // From a phone Activity or Service, get a context for the AI glasses.
+        ProjectedContext.createProjectedDeviceContext(context)
+    } catch (e: IllegalStateException) {
+        // Projected device was not found or connected.
+        null
```
kkuan2011, line 52: may want to log an error here?

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
```diff
@@ -0,0 +1,141 @@
...
+ * Demonstrates how to capture an image using the AI glasses' camera.
+ */
+@androidx.annotation.OptIn(ExperimentalCamera2Interop::class)
+@OptIn(ExperimentalProjectedApi::class)
+// [START androidxr_projected_camera_capture]
+private fun startCameraOnGlasses(activity: ComponentActivity) {
+    // 1. Get the CameraProvider using the projected context.
+    // When using the projected context, DEFAULT_BACK_CAMERA maps to the AI glasses' camera.
+    val projectedContext = ProjectedContext.createProjectedDeviceContext(activity)
+    val cameraProviderFuture = ProcessCameraProvider.getInstance(projectedContext)
```
kkuan2011, line 90: Just wanted to make sure we're handling the null case in the code here.

If projectedContext is null, will this always return a non-null object? could cameraProviderFuture be null?

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Demonstrates how to obtain a context for the host device (phone)
+ * from the projected device (AI glasses).
+ */
+// [START androidxr_projected_context_get_host]
+@OptIn(ExperimentalProjectedApi::class)
+private fun getPhoneContext(activity: ComponentActivity): Context {
+    // From an AI glasses Activity, get a context for the phone.
+    val phoneContext = ProjectedContext.createHostDeviceContext(activity)
```
kkuan2011: nit: should we make this more concise into 1 line?

return ProjectedContext.createHost...

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
```diff
@@ -0,0 +1,141 @@
...
+            .setResolutionSelector(resolutionSelector)
+            .build()
+
+        try {
+            // Unbind use cases before rebinding.
+            cameraProvider.unbindAll()
+
+            // Bind use cases to camera using the Activity as the LifecycleOwner.
+            cameraProvider.bindToLifecycle(
+                activity as LifecycleOwner,
```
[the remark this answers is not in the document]
trambui09: done

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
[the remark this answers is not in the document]
trambui09: done!

trambui09: Yes, thanks for the catch!

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
```diff
@@ -0,0 +1,141 @@
...
+ */
+// [START androidxr_projected_context_get_host]
+@OptIn(ExperimentalProjectedApi::class)
+private fun getPhoneContext(activity: ComponentActivity): Context {
+    // From an AI glasses Activity, get a context for the phone.
+    val phoneContext = ProjectedContext.createHostDeviceContext(activity)
+
+    // Now use phoneContext to access the phone's hardware,
+    // such as sensors or location services.
+    return phoneContext
```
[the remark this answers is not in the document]
trambui09: done

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
trambui09: done!

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
trambui09: I think rememberLauncherForActivityResult cannot be used in onCreate; it can only be used inside a @Composable function. The registerForActivityResult ensures the launcher is registered before the Activity starts.

xr/src/main/java/com/example/xr/projected/ProjectedHardware.kt
trambui09: createProjectedDeviceContext() throws an IllegalStateException if the projected device isn't found. I'll edit the snippets to catch this. I don't think cameraProviderFuture can be null.

xr/src/main/java/com/example/xr/projected/GlassesMainActivity.kt
```diff
@@ -49,6 +55,19 @@ class GlassesMainActivity : ComponentActivity() {
...
+            if (results[Manifest.permission.CAMERA] == true) {
+                // Permission granted, initialize the session/features.
+                initializeGlassesFeatures()
+            } else {
+                // Handle permission denial.
```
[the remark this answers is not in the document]
robclifford27: Unsure if this comment is really required or not without more context. But assuming that the glasses functionality would be severely hampered if the user refuses the camera permission, remediation steps would probably be necessary here.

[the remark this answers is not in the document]
trambui09: Thanks, Rob. I added a simple handling of the permission denial case via the HomeScreen to show a rationale card and a retry button.

PR #827, opened 2026-03-05
[Merge it on 11th March,2026] FlexBox Snippet migration from DSL to function based config

srikrishnasakunia:
Updated the code snippets for FlexBox to migrate from DSL to function based config for FlexBox.


kkuan2011 (commented):
I think the build is failing, is it related to the compose-latest version bump?

PR #857, opened 2026-03-30
Fix spacing of Gemini live snippet

kkuan2011:
So it matches the spacing of the handler function snippet here: https://developer.android.com/ai/gemini/live#function


lethargicpanda (approved):
LGTM

yrezgui (approved):

PR #867, opened 2026-04-09
Added snippets for verified email feature via Credential Manager API

niharika2810:


identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
```diff
@@ -0,0 +1,243 @@
...
+    // Placeholder for generating nonce
+    fun generateSecureRandomNonce(byteLength: Int = 32): String {
+        val randomBytes = ByteArray(byteLength)
+        SecureRandom().nextBytes(randomBytes)
+        return Base64.encodeToString(
+            randomBytes,
+            Base64.NO_WRAP or Base64.URL_SAFE or Base64.NO_PADDING
+        )
+    }
+    // [END android_identity_create_nonce]
```
yrezgui: There's no start for `android_identity_create_nonce`

yrezgui (changes requested):

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
```diff
@@ -0,0 +1,242 @@
...
+        // [END android_credentials_create_account_verified_email_client]
+    }
+    /**
+     * A placeholder function for server-side validation of the digital credential response.
+     *
+     * @param responseJsonString The raw digital credential response JSON.
+     * @param nonce The original nonce used in the OpenID4VP request.
+     * @return A placeholder server response.
+     */
+    private fun createAccountWithVerifiedCredentials(responseJsonString: String, nonce: String): ServerResponse {
```
cy245, line 240: As this is just a snippets file we don't need to fully define functions where the code isn't used in DAC. Maybe we can simplify this and other placeholder functions (i.e. anything that doesn't have start/end tags) to return dummy values?

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Constructs an OpenID4VP request for a UserInfoCredential, sends it to the
+     * CredentialManager, and extracts the verified email and user profile claims
+     * from the returned SD-JWT.
+     *
+     * @param activity The current Activity context.
+     */
+    @OptIn(ExperimentalDigitalCredentialApi::class)
+    suspend fun getVerifiedUserInfo(activity: Activity) {
+        // [START android_credentials_get_verified_user_info_request]
```
cy245: The prefix we've been using is actually [android_identity_...]. Please update here and below

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
```diff
@@ -0,0 +1,242 @@
...
+            val result = credentialManager.getCredential(activity, request)
+
+            when (val credential = result.credential) {
+                is DigitalCredential -> {
+                    val responseJsonString = credential.credentialJson
+
+                    // Successfully received digital credential response.
+
+                    // Next, parse this response and send it to your server.
+                    // [START_EXCLUDE]
```
cy245, line 105: Include prefix? (and end tag too)

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
niharika2810: Hey I believe, nothing harm here to keep them. There were unnecessary updations in the code which I wanted to avoid and to visualize code block as is, trying to keep it simple as per the instruction given

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
niharika2810: Oh I didnt know. Updating it.

identity/credentialmanager/src/main/java/com/example/identity/credentialmanager/VerifiedEmailFunctions.kt
niharika2810: This has end exclude already and this tag excude doesnt need prefix. https://developer.android.com/dac-internal/cct/standard/github/add-snippets#ignore

cy245 (approved):

yrezgui (approved):

PR #881, opened 2026-04-22
Telecom sanitization

natarajkr007:
sanitized the code by addressing comments from PR 876

fixed syntax of snippet's region tags.


telecom/dialer/src/main/AndroidManifest.xml
```diff
@@ -31,7 +31,6 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
-            android:label="@string/app_name"
```
kkuan2011, line 34: why remove this line?

telecom/src/main/java/com/example/telecom/VoipCallActivity.kt
```diff
@@ -37,16 +39,23 @@ class VoipCallbackActivity : ComponentActivity() {
...
+                Box(
+                    modifier = Modifier
+                        .fillMaxSize()
+                        .padding(innerPadding),
+                    contentAlignment = Alignment.Center
+                ) {
+                    Text(
+                        text = "Telecom VoIP snippets",
+                        modifier = Modifier.padding(innerPadding)
+                    )
```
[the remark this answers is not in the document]
kkuan2011: please fix

kkuan2011 (commented):
Thanks for the updates! I think we do want to optimize for faster builds, since a lot of people are working on the snippets repo now, so I do prefer a single module as well for telecom

telecom/src/main/java/com/example/telecom/VoipCallActivity.kt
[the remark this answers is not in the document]
natarajkr007: fixed.

telecom/dialer/src/main/AndroidManifest.xml
natarajkr007: redundant label. Already present in application tag.

natarajkr007:
Merged the two seperate modules in to one to optimize build times.

telecom/.gitignore
```diff
@@ -1,2 +1,3 @@
 /build
 /src/androidTest/
```
kkuan2011: For lines 2 and 3, I don't think we want to ignore these files in the src folders, right? Even if it's not needed now, someone in the future may add some files here and may wonder why it's not getting picked up for their commit.

kkuan2011 (approved):

PR #886, opened 2026-04-23
Identity add issuance attestation snippets

cy245:


kkuan2011 (approved):
Looks good, but would be good to get an identity teammate to review too :)

cy245:
> Looks good, but would be good to get an identity teammate to review too :)

Already approved in the CL, just transcribing here

PR #887, opened 2026-04-27
Update dependencies

MagicalMeghan:
Update dependencies for Compose 1.11


kul3r4 (approved):
LGTM for updating Wear dependenciws

gradle.properties
```diff
@@ -18,13 +18,18 @@ org.gradle.configureondemand=true
 org.gradle.caching=true
 org.gradle.parallel=true
 
+# AndroidX package structure to make it clearer which packages are bundled with the
+# Android operating system, and which are packaged with your app's APK
+# https://developer.android.com/topic/libraries/support-library/androidx-rn
+android.useAndroidX=true
```
kkuan2011: are lines 21-24 needed? I thought useAndroidX is true by default now?

gradle.properties
```diff
@@ -18,13 +18,18 @@ org.gradle.configureondemand=true
...
 # Enable R8 full mode.
 android.enableR8.fullMode=true
 
-android.newDsl=false
+android.newDsl=true
 # Use an AndroidX snapshot build.
 # https://androidx.dev/snapshots/builds
-# snapshotVersion=14793336
\ No newline at end of file
+snapshotVersion=14793336
```
kkuan2011: I think Dereck added this as an example in a comment, but I don't think we need to use a snapshot build for anything right now, unless you're aware of any code that requires it?

gradle.properties
```diff
@@ -18,13 +18,18 @@ org.gradle.configureondemand=true
...
+android.useAndroidX=true
+
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
 
 # Enable R8 full mode.
 android.enableR8.fullMode=true
 
-android.newDsl=false
+android.newDsl=true
```
kkuan2011: since this is true by default, you can remove it now 
https://developer.android.com/build/releases/agp-9-0-0-release-notes#android-gradle-plugin-behavior-changes

gradle/libs.versions.toml
```diff
@@ -56,64 +56,63 @@ datastoreCore = "1.2.1"
...
 datastorePreferencesRxjava3 = "1.2.1"
 firebase-bom = "34.10.0"
-glide = "1.0.0-beta08"
+glide = "1.0.0-beta09"
 google-ar-core = "1.53.0"
 google-maps = "20.0.0"
-gradle-versions = "0.53.0"
-guava = "33.5.0-jre"
-guava-android = "33.5.0-jre"
-health-connect = "1.1.0-alpha11"
```
kkuan2011, line 60: I think health-connect accidentally got deleted, which is causing the spotless check + build to fail, can you add it back? :)

kkuan2011 (commented):
looks good after these comments are addressed!

gradle.properties
MagicalMeghan: agreed to take this out esp for main branch

gradle/libs.versions.toml
```diff
@@ -56,64 +56,64 @@ datastoreCore = "1.2.1"
...
 reactive-streams = "1.0.4"
 recyclerview = "1.4.0"
 registryDigitalCredentials = "1.0.0-alpha04"
 robolectric = "4.16.1"
 roborazzi = "1.59.0"
-spotless = "8.3.0"
-targetSdk = "37"
-tiles = "1.5.0"
+spotless = "8.4.0"
+targetSdk = "36"
```
kkuan2011: should target sdk be left at 37?

gradle.properties
```diff
@@ -18,13 +18,17 @@ org.gradle.configureondemand=true
 org.gradle.caching=true
 org.gradle.parallel=true
 
+# AndroidX package structure to make it clearer which packages are bundled with the
+# Android operating system, and which are packaged with your app's APK
+# https://developer.android.com/topic/libraries/support-library/androidx-rn
+android.useAndroidX=true
```
kkuan2011: can we remove this? I think this is default = true

kmp/shared/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ kotlin {
     // Target declarations - add or remove as needed below. These define
     // which platforms this KMP module supports.
     // See: https://kotlinlang.org/docs/multiplatform-discover-project.html#targets
-    androidLibrary {
+    android {
```
kkuan2011, line 11: if this is a shared KMP library, should we keep it as androidLibrary?

That's how it appears on the Kotlin docs: https://kotlinlang.org/docs/multiplatform/multiplatform-project-agp-9-migration.html#configure-the-shared-module-to-use-the-android-kmp-library-plugin

MagicalMeghan: this is what the lint error says when i try to add androidLibrary:

'fun KotlinMultiplatformExtension.androidLibrary(configure: Action<DeprecatedKotlinMultiplatformAndroidLibraryTarget>): Unit' is deprecated. The 'androidLibrary' block is deprecated. Please use 'android' instead.

kkuan2011: Ah okay I see the note now in the plugin docs: https://developer.android.com/kotlin/multiplatform/plugin#migrate

Never mind, change looks correct, thank you for confirming!

kkuan2011 (approved):

gradle.properties
```diff
@@ -24,7 +24,6 @@ kotlin.code.style=official
 # Enable R8 full mode.
 android.enableR8.fullMode=true
 
-android.newDsl=false
```
raystatic, line 27: This was added earlier to opt out of new DSL introduced by AGP 9.0. 
Are all the snippets compiling properly?

https://github.com/android/snippets/pull/775/changes#r2819103212

kkuan2011: Yes looks like the build is successful. I believe the opt-out was because the protobuf library wasn't compatible with AGP9 but in the latest release it is (https://github.com/google/protobuf-gradle-plugin/releases) so we can remove this now.

raystatic (approved):

yrezgui (approved):

xr/src/main/java/com/example/xr/scenecore/SpatialAudio.kt
```diff
@@ -158,7 +158,7 @@ private fun playSpatialAudioAtEntityAmbionics(session: Session, appContext: Cont
 @OptIn(UnstableApi::class)
 private fun detectSupport(context: Context) {
     // [START androidxr_scenecore_dolby_detect_support]
-    val audioCapabilities = AudioCapabilities.getCapabilities(context, androidx.media3.common.AudioAttributes.DEFAULT, null)
+    val audioCapabilities = AudioCapabilities.getCapabilities(context, androidx.media3.common.AudioAttributes.DEFAULT, null, emptyList())
```
raystatic, line 161: @devbridie replaced the depracated method as asked. Please review.
https://b.corp.google.com/issues/512091859#comment3

devbridie (approved):

PR #889, opened 2026-04-29
Add snippet for SpatialGltfModel

amyZepp:


xr/src/main/java/com/example/xr/compose/SpatialGltfModel.kt
```diff
@@ -0,0 +1,135 @@
...
+    val entityNodes = modelState.nodes
+
+    // Find a specific node by name to apply modifications, such as material overrides.
+    val node = entityNodes.find { it.name == "node_name" }
+
+    // Maintain a reference to the custom material to avoid re-creating it on every recomposition.
+    var pbrMaterial by remember { mutableStateOf<KhronosPbrMaterial?>(null) }
+
+    // Create and apply the custom material once the session is ready and the target node is available.
+    LaunchedEffect( node) {
```
devbridie: suspicious spacing

xr/src/main/java/com/example/xr/compose/SpatialGltfModel.kt
```diff
@@ -0,0 +1,135 @@
...
+        try {
```
devbridie: Do we expect to fail often? It looks rather straightforward, so I would remove unless we're expecting something unexpected to happen often - we can fail fast here

xr/src/main/java/com/example/xr/compose/SpatialGltfModel.kt
amyZepp: updated

xr/src/main/java/com/example/xr/compose/SpatialGltfModel.kt
amyZepp: removed

devbridie (approved):

PR #891, opened 2026-04-30
Update compileSdk and targetSdk to 37 to support Android Studio Quail and the upcoming Compose release

devbridie:


yrezgui (approved):

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/DeviceScanActivity.kt
kkuan2011: should we explicitly define this val as nullable because BluetoothAdapter is a java class and the method won't have the nullable annotation?

kkuan2011 (approved):
LGTM after comment is addressed

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/DeviceScanActivity.kt
devbridie: done!

riggaroo (approved):

PR #902, opened 2026-05-06
Migrate Health connect documentation snippets

shounbi:
Code snippets for [health Connect site](https://developer.android.com/health-and-fitness/health-connect)


healthconnect/src/main/java/com/example/healthconnect/MyWearableService.kt
```diff
@@ -41,14 +55,157 @@ class MyWearableService : CompanionDeviceService() {
 
         serviceScope.launch {
             val granted = healthConnectClient?.permissionController?.getGrantedPermissions()
-            // ... set up GATT and subscribe ...
+
+            // 1. Check permissions ONCE when the device connects
+            if (granted?.contains(HealthPermission.getWritePermission(HeartRateRecord::class)) == true) {
```
kkuan2011: do you need "== true"? I think it would be enough to do the if check on the contains() call.

healthconnect/src/main/java/com/example/healthconnect/MyWearableService.kt
```diff
@@ -41,14 +55,157 @@ class MyWearableService : CompanionDeviceService() {
...
+                TimeRangeFilter.between(startTime, endTime)
+            )
+        )
+
+        val exerciseRecord = readResponse.records.firstOrNull() ?: return
+        val recordId = exerciseRecord.metadata.id
+
+        // 3. Read the specific record to check for the route
+        val sessionResponse = client.readRecord(ExerciseSessionRecord::class, recordId)
+        val exerciseSessionRecord = sessionResponse.record
```
kkuan2011: nit: would it be easier to declare val routeResult here? since it doesn't look like you need exerciseSessionRecord in a val by itself

val routeResult = sessionResponse.record.exerciseRouteResult

when (routeResult) {
...
}

healthconnect/src/main/java/com/example/healthconnect/MyWearableService.kt
```diff
@@ -41,14 +55,157 @@ class MyWearableService : CompanionDeviceService() {
...
+                ExerciseSessionRecord::class,
+                TimeRangeFilter.between(startTime, endTime)
+            )
+        )
+
+        val exerciseRecord = readResponse.records.firstOrNull() ?: return
+        val recordId = exerciseRecord.metadata.id
+
+        // 3. Read the specific record to check for the route
+        val sessionResponse = client.readRecord(ExerciseSessionRecord::class, recordId)
```
kkuan2011, line 118: could this be null?

kkuan2011 (approved):

healthconnect/src/main/java/com/example/healthconnect/MyWearableService.kt
shounbi: readRecord returns a non-nullable response object, so the result itself won't be null.

yrezgui (approved):

PR #915, opened 2026-05-13
Add health connect snippets

shounbi:
Add snippets for Health Connect site


healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
         }
     }
 
+    suspend fun  createSetOfPermissionSleep(
```
yrezgui:
```suggestion
    fun createSetOfPermissionSleep(
```

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                HealthPermission.getReadPermission(OxygenSaturationRecord::class),
+                HealthPermission.getWritePermission(OxygenSaturationRecord::class),
+                HealthPermission.getReadPermission(RespiratoryRateRecord::class),
+                HealthPermission.getWritePermission(RespiratoryRateRecord::class)
+            )
+            // [END android_healthconnect_create_set_of_permissions_sleep]
+        return permissions
+    }
+
+    suspend fun  createSetOfPermissionVitals(
```
yrezgui:
```suggestion
    fun createSetOfPermissionVitals(
```

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                HealthPermission.getReadPermission(HeartRateRecord::class),
+                HealthPermission.getWritePermission(HeartRateRecord::class),
+                HealthPermission.getReadPermission(BodyTemperatureRecord::class),
+                HealthPermission.getWritePermission(BodyTemperatureRecord::class)
+            )
+        // [END android_healthconnect_create_set_of_permissions_vitals]
+        return permissions
+    }
+
+    suspend fun  createSetOfPermissionWorkout(
```
yrezgui:
```suggestion
    fun createSetOfPermissionWorkout(
```

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+        // [START android_healthconnect_check_permission_granted]
+        if (!granted.containsAll(permissions)) {
+            // Check if required permissions are not granted, and return
+        }
+        // Permissions already granted; proceed with inserting or reading data
+        // [START android_healthconnect_check_permission_granted]
+        return permissions
+    }
+
+    suspend fun  createSetOfPermissionGetStarted(
```
yrezgui:
```suggestion
    fun createSetOfPermissionGetStarted(
```

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                HealthPermission.getReadPermission(HeartRateRecord::class),
+                HealthPermission.getWritePermission(HeartRateRecord::class),
+                HealthPermission.getReadPermission(StepsRecord::class),
+                HealthPermission.getWritePermission(StepsRecord::class)
+            )
+        // [END android_healthconnect_create_set_of_permissions_user]
+        return permissions
+    }
+
+    suspend fun  createPermissionStep(
```
yrezgui:
```suggestion
    fun createPermissionStep(
```

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+        val permissions =
+            setOf(
+                HealthPermission.getReadPermission(StepsRecord::class),
+                HealthPermission.getWritePermission(StepsRecord::class)
+            )
+        // [END android_healthconnect_create_set_of_permissions_step]
+        return permissions
+    }
+
+    suspend fun  createPermissionHeartRate(): Set<String> {
```
yrezgui:
```suggestion
    fun createPermissionHeartRate(): Set<String> {
```

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
         }
     }
 
+    suspend fun  createSetOfPermissionSleep(
+        healthConnectClient: HealthConnectClient,
+        record: StepsRecord
```
yrezgui: Parameters not needed

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                HealthPermission.getReadPermission(RespiratoryRateRecord::class),
+                HealthPermission.getWritePermission(RespiratoryRateRecord::class)
+            )
+            // [END android_healthconnect_create_set_of_permissions_sleep]
+        return permissions
+    }
+
+    suspend fun  createSetOfPermissionVitals(
+        healthConnectClient: HealthConnectClient,
+        record: StepsRecord
```
yrezgui: Parameters not needed

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                HealthPermission.getReadPermission(BodyTemperatureRecord::class),
+                HealthPermission.getWritePermission(BodyTemperatureRecord::class)
+            )
+        // [END android_healthconnect_create_set_of_permissions_vitals]
+        return permissions
+    }
+
+    suspend fun  createSetOfPermissionWorkout(
+        healthConnectClient: HealthConnectClient,
+        record: StepsRecord
```
yrezgui: Parameter not needed

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+            // Check if required permissions are not granted, and return
+        }
+        // Permissions already granted; proceed with inserting or reading data
+        // [START android_healthconnect_check_permission_granted]
+        return permissions
+    }
+
+    suspend fun  createSetOfPermissionGetStarted(
+        healthConnectClient: HealthConnectClient,
+        record: StepsRecord
```
yrezgui: Parameters not needed

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                HealthPermission.getReadPermission(StepsRecord::class),
+                HealthPermission.getWritePermission(StepsRecord::class)
+            )
+        // [END android_healthconnect_create_set_of_permissions_user]
+        return permissions
+    }
+
+    suspend fun  createPermissionStep(
+        healthConnectClient: HealthConnectClient,
+        record: StepsRecord
```
yrezgui: Parameters not needed

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +272,247 @@ class HealthConnectManager(
...
+                device = Device(type = Device.Companion.TYPE_PHONE)
+            )
+        )
+        val insertedPlannedExerciseSessions =
+            healthConnectClient.insertRecords(listOf(plannedExerciseSessionRecord)).recordIdsList
+        val insertedPlannedExerciseSessionId = insertedPlannedExerciseSessions.first()
+        // [END android_healthconnect_check_permission_planned_excerise_session]
+    }
+
+    @SuppressLint("RestrictedApi")
```
yrezgui, line 500: Why is this required?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -978,4 +1237,45 @@ class HealthConnectManager(
         healthConnectClient.insertRecords(listOf(session))
         // [END android_healthconnect_write_exercise_route]
     }
+
+    @SuppressLint("RestrictedApi")
```
yrezgui, line 1247: Why is this required?

yrezgui (changes requested):
You need to apply spotless

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: I added it to fix the lint error I was getting `ExerciseSessionRecord can only be called from within the same library (androidx.health.connect:connect-client)`

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: I added it to fix the lint error I was getting `
ExerciseSessionRecord can only be called from within the same library (androidx.health.connect:connect-client)`

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: Suspend function 'suspend fun getGrantedPermissions(): Set<String>' can only be called from a coroutine or another suspend function.

shounbi:
> You need to apply spotless

I have ran spotless command ( ./gradlew spotlessApply)

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
kkuan2011: how about this pending comment?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +274,234 @@ class HealthConnectManager(
...
+            title = "Run at lake",
+            notes = null,
+            metadata = Metadata(
+                device = Device(type = Device.Companion.TYPE_PHONE),
+            ),
+            startZoneOffset = null,
+            endZoneOffset = null,
+        )
+        val insertedPlannedExerciseSessions =
+            healthConnectClient.insertRecords(listOf(plannedExerciseSessionRecord)).recordIdsList
```
kkuan2011: do you need to check that insertRecords() was successful before calling recordIdsList? and getting the first index from it?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +274,234 @@ class HealthConnectManager(
...
+        // [END android_healthconnect_check_permission_planned_exercise_session]
+    }
+
+    @SuppressLint("RestrictedApi")
+    suspend fun  insertRecords(
+        healthConnectClient: HealthConnectClient,
+        startTime: Instant
+    ){
+        // [START android_healthconnect_create_insert_record]
+        val sessionId = UUID.randomUUID().toString()
```
kkuan2011: nit: does this need to be used anywhere?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
kkuan2011: same here, what do you think of this pending comment?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
kkuan2011: Is this okay for an external dev to call then? or is it an internal API only meant for being called within the library? And same question for the other times you suppress the lint error below too.

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -978,4 +1226,173 @@ class HealthConnectManager(
...
+        )
+        val insertedHeartRateRecords = healthConnectClient.insertRecords(listOf(heartRateRecord))
+        // [END android_healthconnect_insert_segment_exercise_session_with_handler]
+    }
+
+    suspend fun checkPermissionsAndRun(
+        healthConnectClient: HealthConnectClient,
+        launcher: ActivityResultLauncher<Set<String>>
+    ) {
+        val PERMISSIONS = setOf(
```
kkuan2011: I think this just can be lowercase "permissions" because it's not a constant or top level val https://kotlinlang.org/docs/coding-conventions.html#property-names

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -978,4 +1226,173 @@ class HealthConnectManager(
...
+            samples = samples,
+            metadata = Metadata(
+                device = Device(type = Device.Companion.TYPE_WATCH)
+            )
+        )
+        val insertedHeartRateRecords = healthConnectClient.insertRecords(listOf(heartRateRecord))
+        // [END android_healthconnect_insert_segment_exercise_session_with_handler]
+    }
+
+    suspend fun checkPermissionsAndRun(
```
kkuan2011: I think this isn't quite the full picture of how we ideally show developers how to request permissions because the "Permissions granted, proceed with inserting/reading data" should be part of the registerForActivityResult callback. What do you think about showing this code within the context of an activity?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: No it's not, i'll take it out.

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: Thanks for pointing this out! insertRecords() throws an exception on failure rather than returning an empty response, so execution would stop before reaching .first(). That said, you're absolutely right that we should handle failures gracefully. I’ve wrapped it in a try-catch block to handle exceptions and switched to firstOrNull() for safety.

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: Good catch. You're completely right, the previous snippet didn’t show how the asynchronous flow actually behaves, leaving the callback completely missing.

Since this page uses Jetpack Compose, I've moved the permission flow directly into the context of `HealthConnectScreen` using `rememberLauncherForActivityResult`.

Now, developers can clearly see the contract registration, the explicit callback handling, and how the runtime permission check guards the button actions.

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: Yes, it is completely safe and intended for external developers. The RestrictedApi lint error here is a known bug in Google’s Jetpack Health Connect library compilation, not an intentional restriction on external developers.

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -115,6 +120,27 @@ fun HealthConnectScreen(modifier: Modifier) {
         healthConnectClient?.let { HealthConnectManager(it) }
     }
 
+    // [START android_healthconnect_check_permission_launcher]
+    val permissions = remember {
```
kkuan2011: I don't think we need "remember {}" here because it's a static list, can we just have permissions = setOf(...)?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -135,10 +161,15 @@ fun HealthConnectScreen(modifier: Modifier) {
...
                         coroutineScope.launch {
-                            val startTime = Instant.now().minusSeconds(3600)
-                            val endTime = Instant.now()
-                            manager.insertSteps(startTime, endTime)
-                            snackbarHostState.showSnackbar("Steps inserted!")
+                            val granted = healthConnectClient?.permissionController?.getGrantedPermissions()
+                            if (granted?.containsAll(permissions) == true) {
+                                val startTime = Instant.now().minusSeconds(3600)
+                                val endTime = Instant.now()
+                                manager.insertSteps(startTime, endTime)
```
kkuan2011: does this insertSteps() call need to be on an I/O thread? because I thought coroutineScope uses the main dispatcher by default

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +274,251 @@ class HealthConnectManager(
         }
     }
 
+    fun  createSetOfPermissionSleep(): Set<String> {
```
kkuan2011, line 277: return value can be Unit for these createPermission..() methods in this file right?

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
```diff
@@ -254,6 +274,251 @@ class HealthConnectManager(
...
+        val permissions =
+            setOf(
+                HealthPermission.getReadPermission(HeartRateRecord::class),
+                HealthPermission.getWritePermission(HeartRateRecord::class)
+            )
+        // [END android_healthconnect_create_set_of_permissions_heart_rate]
+        return permissions
+    }
+
+    suspend fun  createPermissionSleepSession(): Set<String> {
```
kkuan2011: for the methods in this file, can you double check that you only add suspend onto the functions that require it?  for example, I think suspend can be removed here and we can return Unit for the method

healthconnect/src/main/java/com/example/healthconnect/HealthConnectManager.kt
shounbi: No, It cannot be Unit if you actually want to use these permissions later in your code. Right now It builds the set and returns it so that another part of your app (like your permission request launcher) can actually use it to ask the user for access.

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -134,11 +160,28 @@ fun HealthConnectScreen(modifier: Modifier) {
...
                 item {
                     Button(onClick = {
+                        // Starts on Main thread by default
                         coroutineScope.launch {
-                            val startTime = Instant.now().minusSeconds(3600)
-                            val endTime = Instant.now()
-                            manager.insertSteps(startTime, endTime)
-                            snackbarHostState.showSnackbar("Steps inserted!")
+                            // 1. Check permissions on IO thread, but await the result here on Main
+                            val granted = withContext(Dispatchers.IO) {
```
kkuan2011: I believe you don't need to wrap this call in the withContext(Dispatchers.IO) block:
healthConnectClient?.permissionController?.getGrantedPermissions()

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -150,13 +193,29 @@ fun HealthConnectScreen(modifier: Modifier) {
                         modifier = Modifier.padding(top = 8.dp),
                         onClick = {
                             coroutineScope.launch {
-                                val startTime = Instant.now().minus(Duration.ofDays(1))
-                                val endTime = Instant.now()
+                                // Check permissions on the I/O thread pool
+                                val granted = withContext(Dispatchers.IO) {
```
kkuan2011: same here, I believe you don't need to wrap this call in a withContext(Dispatchers.IO) block

kkuan2011 (approved):

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -134,11 +160,27 @@ fun HealthConnectScreen(modifier: Modifier) {
             if (manager != null) {
                 item {
                     Button(onClick = {
+                        // Starts on Main thread by default
                         coroutineScope.launch {
-                            val startTime = Instant.now().minusSeconds(3600)
-                            val endTime = Instant.now()
-                            manager.insertSteps(startTime, endTime)
-                            snackbarHostState.showSnackbar("Steps inserted!")
+                            // 1. Check permissions on IO thread, but await the result here on Main
```
kkuan2011: nit: update comment

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -134,11 +160,27 @@ fun HealthConnectScreen(modifier: Modifier) {
...
+                            val granted = healthConnectClient?.permissionController?.getGrantedPermissions()
+
+
```
kkuan2011: nit: remove extra blank line

healthconnect/src/main/java/com/example/healthconnect/HealthConnectActivity.kt
```diff
@@ -150,13 +192,27 @@ fun HealthConnectScreen(modifier: Modifier) {
                         modifier = Modifier.padding(top = 8.dp),
                         onClick = {
                             coroutineScope.launch {
-                                val startTime = Instant.now().minus(Duration.ofDays(1))
-                                val endTime = Instant.now()
+                                // Check permissions on the I/O thread pool
```
kkuan2011: update comment

kkuan2011 (approved):

kkuan2011:
I'm going to merge now, but can you address the nits in a follow up CL?

kkuan2011:
actually I'm not able to merge, I think Yacine needs to change his review status?

yrezgui (approved):

PR #920, opened 2026-05-14
Add memory leak examples for Android documentation

sudeshim3:
* Add Pattern 1: UI Objects Retained Beyond Lifecycle examples
* Add Pattern 2: Background Work Outlives the UI examples
* Add Pattern 3: Unreleased External Registrations examples
* Include Compose and Fragment-specific memory management best practices
* Ensure snippets are idiomatic and comply with Spotless formatting

These examples illustrate common architectural pitfalls and their corresponding lifecycle-aware resolutions using Coroutines, Flows, and DisposableEffect.


misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+// [END android_memory_leak_repository_callback_with_leak]
+
+// [START android_memory_leak_repository_callback_recommended]
+class UserRepositoryRecommended {
+    suspend fun fetchUser(): User {
+        return api.getUser()
+    }
+}
+
+/* In the Fragment/UI layer:
```
AliceYuan: Instead of showing commented out code you should show the actual compileable code, also I think you could show this occuring within a view model instead of calling from activity to avoid other code smells.

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+viewLifecycleOwner.lifecycleScope.launch {
+    val user = repository.fetchUser()
+    binding.name.text = user.name
+}
+*/
+// [END android_memory_leak_repository_callback_recommended]
```
AliceYuan: Recommend creating 2 code snippets here if the intention is to show repository.fetchUser() getting called. What I mean is it could look something like:
```
// [START android_memory_leak_repository_callback_recommended_pt1]
class UserRepositoryRecommended {
    suspend fun fetchUser(): User {
        return api.getUser()
    }
}
// [END android_memory_leak_repository_callback_recommended_pt1]

class UserViewModel(private val repository: UserRepositoryRecommended) : ViewModel() {

    private val _userState = MutableStateFlow<User?>(null)
    val userState: StateFlow<User?> = _userState.asStateFlow()

// [START android_memory_leak_repository_callback_recommended_pt2]
    fun loadUser() {
        // viewModelScope automatically cancels if the user leaves the screen
        viewModelScope.launch {
                val user = repository.fetchUser()
                _userState.value = user
        }
    }
// [END android_memory_leak_repository_callback_recommended_pt2]
}
```

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+
+// [START android_memory_leak_delayed_work_with_leak]
+// Singleton scope accepts a UI-bound callback
+object UserRepositoryDelayedWithLeak {
+    private val repositoryScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+
+    // Accepts a callback that might capture a destroyed UI Context
+    fun fetchUserDataWithDelay(onComplete: (String) -> Unit) {
+        repositoryScope.launch {
+            delay(5_000)
```
AliceYuan: suggest adding some explanation that the delay is to emulate network or processing.

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+// [END android_memory_leak_delayed_work_with_leak]
+
+// [START android_memory_leak_delayed_work_recommended]
+//  Expose data as a Flow and let the UI handle the lifecycle scope
+object UserRepositoryDelayedRecommended {
+    // A clean, stateless flow with no callback parameters
+    fun getUserData(): Flow<String> = flow {
+        delay(5_000)
```
AliceYuan: suggest adding some explanation that the delay is to emulate network or processing.

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+    fun fetchUser(callback: (User) -> Unit) {
+        // The repository retains the callback beyond the view lifecycle.
+        listener = callback
+        api.getUser { user ->
+            listener?.invoke(user)
+        }
+    }
+}
+
+/* In the Fragment/UI layer:
```
AliceYuan: same recommendation to actually create the Activity (or ViewModel) that calls the repository.

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+class MainActivityRecommended : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        // Automatically cancels collection and releases MainActivity when destroyed
+        lifecycleScope.launch {
+            UserRepositoryDelayedRecommended.getUserData().collect { data ->
+                title = data
+            }
+        }
```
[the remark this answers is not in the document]
AliceYuan: looks like gemini detected the code smell we're trying to show as an example

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+// [END android_memory_leak_delayed_work_recommended]
+
+// Pattern 3: Example 2 - Fragment view binding is not cleared
```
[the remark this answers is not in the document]
AliceYuan: I think I agree with this assessment actually should we remove? It seems fairly similar as Pattern 1, option2.

I don't think a 2nd example is needed

AliceYuan (approved):
There's still an open question about removing the last view binding example but approving

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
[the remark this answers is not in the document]
sudeshim3: This has been resolved. I am using repeatOnLifecycle(Lifecycle.State.STARTED)

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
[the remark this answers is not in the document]
sudeshim3: Thanks for noticing this. It got added by mistake. Its a unrelated example for Pattern 3 and its already covered in Pattern 1 - Example 1. I will remove this

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
sudeshim3: I removed the delay() call, as it doesn't accurately represent a real world scenario. Instead, I replaced it with a mock API call to show a asynchronous operation is getting called.

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
sudeshim3: Removed delay() call

yrezgui (approved):

kkuan2011 (approved):

misc/src/main/java/com/example/snippets/memoryleaks/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,236 @@
...
+            // network or database call that suspends
+            val user = api.getUser()
+            // If the Activity was destroyed while waiting for the API,
+            // invoking this callback will leak the Activity!
+            onComplete(user.name)
+        }
+    }
+}
+
+class MainActivityWithLeak : AppCompatActivity() {
```
kkuan2011: can you make this ComponentActivity to make the example more compose-friendly? and below as well

compose/snippets/src/main/java/com/example/compose/snippets/performance/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2024 The Android Open Source Project
```
kkuan2011: nit: 2026 and same for next file too

kkuan2011 (approved):

compose/snippets/src/main/java/com/example/compose/snippets/performance/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    Text(text = locationText)
+}
+// [END android_compose_performance_memory_leak_location_with_leak]
+
+@SuppressLint("MissingPermission")
+// [START android_compose_performance_memory_leak_location_recommended]
+@Composable
+fun LocationScreenRecommended(locationManager: LocationManager) {
+    var locationText by remember { mutableStateOf("Locating...") }
```
raystatic, line 54: It is recommended to use [rememberUpdatedState](https://developer.android.com/develop/ui/compose/side-effects?content_ref=reference%20a%20value%20in%20an%20effect%20that%20shouldn%20t%20restart%20if%20the%20value%20changes#rememberupdatedstate) for long-live operations.

compose/snippets/src/main/java/com/example/compose/snippets/performance/MemoryLeakSnippets.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    // DisposableEffect provides an onDispose block for mandatory cleanup
+    DisposableEffect(locationManager) {
```
raystatic, line 57: Also, please check [LifecycleStartEffect](https://developer.android.com/topic/libraries/architecture/lifecycle?content_ref=val%20listener%20locationlistener%20location%20update%20ui%20locationmanager%20requestlocationupdates%20listener%20the%20cleanup%20block%20automatically%20runs%20on%20on_stop%20or%20on%20disposal%20onstopordispose%20locationmanager%20removeupdates%20listener#lifecycle-start)

It shows a location manager example only.

sudeshim3: There are many platform-specific lifecycle handlers that developers should adopt based on their exact requirements. the primary focus of this document is on general memory leak patterns and bringing an understanding of how failing to clean up a UI component causes a leak. Introducing specific lifecycle APIs brings the focus to the example(Location Based app) instead of communicating the concept.
For example: LifecycleStartEffect aggressively tears down on onStop (when the app goes to the background), whereas DisposableEffect keeps a connection alive as long as the component is still in the UI tree. Whether you do it when the app goes to the background, or when your component is getting destroyed, is left for the developer to explore based on their exact needs.

compose/snippets/src/main/java/com/example/compose/snippets/performance/MemoryLeakSnippets.kt
sudeshim3: Sticking to the base remember keeps the example easy to read while still being technically correct. My goal was just to show the setup/teardown mechanism without getting bogged down in advanced Compose state mechanics. It will shift the the readers focus from 'how to clean up resources' to 'how Compose handles stale closures in lambdas'

PR #934, opened 2026-06-01
fix(bluetoothle): fix typo in BluetoothLeService and migrate to snippets

ithinkihaveacat:
- Fix typo 'Return;' -> 'return;' in Java BluetoothLeService.
- Migrate inline code samples from connect-gatt-server.md to snippets.
- Create BluetoothLeService and DeviceControlActivity in Java and Kotlin.
- Add gatt_services_characteristics layout and string resources to enable compilation.

Bug: 359501045


ithinkihaveacat:
These snippets have been ported directly from <https://developer.android.com/develop/connectivity/bluetooth/ble/connect-gatt-server> using a tool I'm testing. 

The comments from gemini-code-assist seem mostly reasonable however I am not a SME and so I'd rather not make any changes if possible.

kkuan2011:
Thanks Michael! Would you mind getting a Bluetooth SME to approve first? Then Yacine/I can approve to merge.

ithinkihaveacat:
Hey @kkuan2011 

> Thanks Michael! Would you mind getting a Bluetooth SME to approve first? Then Yacine/I can approve to merge.

I checked with our Bluetooth SME and made a small change (https://github.com/android/snippets/pull/934/commits/0007881522da9d527ee1742c6fca11c1a75a065a).

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/DeviceControlActivity.kt
```diff
@@ -0,0 +1,172 @@
...
+    private fun makeGattUpdateIntentFilter(): IntentFilter {
+        return IntentFilter().apply {
+            addAction(BluetoothLeService.ACTION_GATT_CONNECTED)
+            addAction(BluetoothLeService.ACTION_GATT_DISCONNECTED)
+        }
+    }
+    // [END android_bluetooth_receiver_lifecycle]
+
+    private fun updateConnectionState(resourceId: Int) {
+        // Dummy implementation
```
kkuan2011: nit: instead of dummy, can we use a more inclusive term like placeholder?

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,172 @@
...
+public class BluetoothLeService extends Service {
+
+    public static final String TAG = "BluetoothLeService";
+
+    private final Binder binder = new LocalBinder();
+    private BluetoothAdapter bluetoothAdapter;
+    private BluetoothGatt bluetoothGatt;
+    private int connectionState;
+
+    // [START android_bluetooth_constants_java]
```
kkuan2011: Can you have 2 start tags right after each other? (Line 34 and line 44) without an end tag in the middle?

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,172 @@
...
+            "com.example.bluetooth.le.ACTION_GATT_CONNECTED";
+    public static final String ACTION_GATT_DISCONNECTED =
+            "com.example.bluetooth.le.ACTION_GATT_DISCONNECTED";
+
+    private static final int STATE_DISCONNECTED = 0;
+    private static final int STATE_CONNECTED = 2;
+    // [END android_bluetooth_constants_java]
+
+
+    // [START android_bluetooth_binder_java]
```
kkuan2011: This snippet looks different than the original one with the Service declaration + binder initialization around it. Is that intended?

class BluetoothLeService extends Service {

    private Binder binder = new LocalBinder();

    @Nullable
    @Override
    public IBinder onBind(Intent intent) {
        return binder;
    }

    class LocalBinder extends Binder {
        public BluetoothLeService getService() {
            return BluetoothLeService.this;
        }
    }
}

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,172 @@
...
+        if (bluetoothAdapter == null) {
+            Log.e(TAG, "Unable to obtain a BluetoothAdapter.");
+            return false;
+        }
+        return true;
+    }
+    // [END android_bluetooth_initialize_java]
+
+    // [START android_bluetooth_connect_simple_java]
+    public boolean simpleConnect(final String address) {
```
kkuan2011: this method name changed (original was connect()), and the paragraph below it in the DAC page references the connect() function so I think we need to keep the original name, or update the markdown text

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,172 @@
...
+            Log.w(TAG, "Device not found with provided address.");
+            return false;
+        }
+        // connect to the GATT server on the device
+        return true;
+    }
+    // [END android_bluetooth_connect_simple_java]
+
+    // [START android_bluetooth_connect_java]
+    @RequiresPermission(Manifest.permission.BLUETOOTH_CONNECT)
```
kkuan2011: Was this annotation in the original code? I can't seem to find it.. but I do understand that we need it for our snippets. I'm wondering if we should put the annotation tag before the start tag.

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public boolean onUnbind(Intent intent) {
+        close();
+        return super.onUnbind(intent);
+    }
+
+    @RequiresPermission(Manifest.permission.BLUETOOTH_CONNECT)
+    private void close() {
+        if (bluetoothGatt == null) {
+            return; // FIXED: lowercase 'return'
```
kkuan2011: comment not needed

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/DeviceControlActivity.kt
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        override fun onServiceDisconnected(componentName: ComponentName) {
+            bluetoothService = null
+        }
+    }
+    // [END android_bluetooth_service_connection]
+
+    // [START android_bluetooth_update_receiver]
```
kkuan2011: it looks like this snippet is longer in the original file with multiple methods include onResume, onPause, etc..

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,172 @@
...
+            return true;
+        } catch (IllegalArgumentException exception) {
+            Log.w(TAG, "Device not found with provided address.  Unable to connect.");
+            return false;
+        }
+    }
+    // [END android_bluetooth_connect_java]
+
+    // [START android_bluetooth_callback_simple_java]
+    private final BluetoothGattCallback simpleBluetoothGattCallback = new BluetoothGattCallback() {
```
kkuan2011: this was a name change from the original code, is that intended?

kkuan2011 (changes requested):
When I do a side by side visual comparison with the original page, I see the snippets have changed 
https://developer.android.com/develop/connectivity/bluetooth/ble/connect-gatt-server 

They omit the Service declaration that wraps a lot of the original snippets, function names have changed (which I understand was to avoid name conflicts within the same file), some global variables also got dropped, annotations got added. Is that what you were expecting?

I do try handling some of these cases in my snippet skill so you may want to try migrating the page with that, or fixing each difference in your current CL by iterating with the agent. It's a tedious process though.. i know :/

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
yrezgui, line 1: Can we drop the Java snippets entirely? With Android now being Compose first, we should reduce the maintenance burden

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
ithinkihaveacat: Yeah, that's probably cleaner. You can click through to the source code to see the full details. Moved (in a few other places as well).

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
ithinkihaveacat: 😬 Oops, removed.

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/DeviceControlActivity.kt
ithinkihaveacat: Done.

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
ithinkihaveacat: Looks weird but you can! e.g. https://github.com/android/snippets/blob/main/compose/snippets/src/main/java/com/example/compose/snippets/touchinput/pointerinput/TapAndPress.kt#L118-L139

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
ithinkihaveacat: No—thanks. Rewrote using nested private object wrappers.

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/DeviceControlActivity.kt
ithinkihaveacat: Fixed.

ithinkihaveacat:
> When I do a side by side visual comparison with the original page, I see the snippets have changed https://developer.android.com/develop/connectivity/bluetooth/ble/connect-gatt-server
>
> [...]
> 
> Is that what you were expecting?

No, and I apologize for missing this, and the other issues you pointed out!

These should now be resolved, although it's quite difficult for anyone to review without the context of a rendered DAC page that incorporates the snippets.

Note: as part of the most recent changes I removed the outer `class` wrappers from these snippets to keep the repo clean. Because the page shows progressive steps of the same class (like `initialize`, `connect`, and `close`), preserving the wrappers in a single file results in conflicting tag exclusions. I think this is the right way to go: keeping them would have forced us to split the code into 5-6 duplicate files in the repo (one per snippet).

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/BluetoothLeService.kt
```diff
@@ -0,0 +1,186 @@
...
+
+private const val TAG = "BluetoothLeService"
+
+// [START android_bluetooth_service_all]
+class BluetoothLeService : Service() {
+
+    private val binder = LocalBinder()
+
+    private var bluetoothAdapter: BluetoothAdapter? = null
+    private var bluetoothGatt: BluetoothGatt? = null
```
kkuan2011, line 44: is ok to omit this from the snippet? 

Original snippet had:

var bluetoothGatt: BluetoothGatt? = null
...

bluetoothGatt = device.connectGatt(this, false, gattCallback)

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/BluetoothLeService.kt
```diff
@@ -0,0 +1,186 @@
...
+    private var connectionState = STATE_DISCONNECTED
+
+    // [START android_bluetooth_callback]
```
kkuan2011: I believe this snippet looks different from the original, and adds discoverServices() call and a @requirespermission annotation - is that ok?

bluetoothle/src/main/java/com/sample/android/bluetoothle/java/BluetoothLeService.java
```diff
@@ -0,0 +1,187 @@
...
+            return false;
+        }
+        return true;
+    }
+    // [END android_bluetooth_initialize_java]
+
+    // [START android_bluetooth_connect_simple_java]
+    // [START_EXCLUDE silent]
+    private class SimplifiedConnect {
+        public boolean connect(final String address) {
```
kkuan2011: For a couple of these snippets, we're excluding the function signature, even though they were shown in the DAC page, is that intended?

kkuan2011 (commented):
I haven't done a full review yet, but since it looks like what will be shown in the resulting DAC page will look different, could you run it by a bluetooth SME one more time? It may be easier to generate a diff report of what is different, so it's easier to see + confirm if those changes are ok. Thank you Michael!

yrezgui (changes requested):

ithinkihaveacat:
Ok, flipping this PR back to ready for review. Thanks for the comments! 

Here's a summary of the PR and the rationale for the (small!) differences between the repository code and the original hardcoded documentation.

### 1. Bug Fixes, Safety & Correctness
*   **Null Safety:** Replaced unsafe null assertions (`!!`) with safe calls (`?.let`) in `DeviceControlActivity.kt` to prevent runtime crashes.
*   **Typo Fix:** Corrected a typo in `DeviceControlActivity.kt` where it referenced a non-existent plural activity class (`DeviceControlsActivity.TAG`) for the log `TAG`.
*   **Control Flow Fix:** Added a missing `return` statement in `BluetoothLeServiceSimple.kt`'s `connect` function to ensure correct compilation.
*   **Lint Compliance:** Resolved a `MissingPermission` lint failure on `gatt.close()` by applying a class-level `@SuppressLint("MissingPermission")` in `BluetoothLeService.kt`. The annotation is placed outside the region tags to keep it hidden from the rendered documentation.
*   **Java Snippets:** Removed the Java implementation entirely (per previous review agreement) to align with Android's Kotlin-first direction and reduce maintenance overhead.

### 2. Structural Changes & Render Differences
To prioritize repository maintainability and avoid duplicating entire Service/Activity files 5-6 times, we consolidated progressive steps of the same class into single helper files (`DeviceControlActivitySimple.kt`, `BluetoothLeServiceSimple.kt`). This results in some minor, intentional differences in the rendered documentation:
*   **Contextual Comments:** Since outer class wrappers were removed to keep the repository clean, we added explicit comments (e.g., `// In DeviceControlActivity`) at the top of the snippets to preserve necessary context for the reader.
*   **Boilerplate Reduction:** Some snippets now focus purely on the functional API calls (e.g., the `connect` function) rather than the surrounding class boilerplate, making them cleaner and more focused.
*   **Collision Resolution:** Naming collisions in helper files were resolved using private inner classes, which are silently excluded from the rendered output using DevSite tags to keep the documentation clean.
*   **Layout Refinement:** In some cases, we reordered elements (like consolidating the companion object and callback) to present a more contiguous and readable flow on the page.

### 3. Rendered Preview
The changes have been verified using an internal staging system. The rendered output matches the original documentation layout while incorporating the safety, correctness, and context improvements noted above. *(Details shared separately).*

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/BluetoothLeService.kt
```diff
@@ -0,0 +1,331 @@
...
+    }
+    // [END android_bluetooth_initialize]
+}
+
+@SuppressLint("MissingPermission")
+private object ConnectGattNamespace {
+    class BluetoothLeService : Service() {
+        // [START android_bluetooth_connect_gatt]
+        // [START_EXCLUDE silent]
+        fun dummy(device: BluetoothDevice, bluetoothGattCallback: BluetoothGattCallback) {
```
kkuan2011: nit: can we call it a different placeholder name?

bluetoothle/src/main/java/com/sample/android/bluetoothle/kotlin/BluetoothLeService.kt
```diff
@@ -0,0 +1,331 @@
...
+    class BluetoothLeService : Service() {
+
+        // [START_EXCLUDE silent]
+        private var bluetoothAdapter: BluetoothAdapter? = null
+        private val bluetoothGattCallback = object : BluetoothGattCallback() {}
+        private val TAG = "BluetoothLeService"
+        // [END_EXCLUDE]
+        // [START_EXCLUDE]
+        override fun onBind(intent: Intent): IBinder? = null
+        fun dummy() {}
```
kkuan2011: same here and check for other occurences in the files too

kkuan2011 (approved):
Looks good after comments are addressed, thank you!!

ithinkihaveacat:
@yrezgui Can you re-review? I've removed the Java snippets as requested.

PR #936, opened 2026-06-04
Mesh Gradients in Compose 1.12

riggaroo:
Adding snippets for https://developer.android.com/develop/ui/compose/graphics/draw/mesh-gradient


gradle/libs.versions.toml
```diff
@@ -144,7 +140,7 @@ androidx-camera-viewfinder-compose = { module = "androidx.camera.viewfinder:view
 androidx-camera-camera2 = { module = "androidx.camera:camera-camera2", version.ref = "androidx-cameraX" }
 androidx-car = { module = "androidx.car.app:app", version.ref = "androidx-car"}
 androidx-compose-animation-graphics = { module = "androidx.compose.animation:animation-graphics", version.ref = "compose-latest" }
-androidx-compose-bom = { module = "androidx.compose:compose-bom", version.ref = "androidx-compose-bom" }
+androidx-compose-bom = { module = "androidx.compose:compose-bom-alpha", version.ref = "androidx-compose-bom" }
```
kkuan2011: is it okay move the whole repo to alpha?

kkuan2011 (commented):
Changes look good, except one question:

gradle/libs.versions.toml
riggaroo: @kkuan2011 no, will leave this as a branch until we are good to go with a stable version :)

kkuan2011: ok sounds good :) I'll assume no action needed for now then

compose/snippets/src/main/java/com/example/compose/snippets/graphics/MeshGradientSnippets.kt
```diff
@@ -0,0 +1,219 @@
...
+
+    val coral = Color(255, 90, 90)
+    val peach = Color(255, 139, 90)
+    val amber = Color(255, 169, 90)
+    val sunshine = Color(255, 212, 90)
+    val indigo = Color(0xFF5856D6)
+    val pink = Color(0xFFFF2D55)
+
+
+    val gradientPainter = remember {
```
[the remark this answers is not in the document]
riggaroo: This is not true. The lambda is only executed in the DrawScope, and accesses the variable there. There is no need for providing a key to remember here and in this case it will only harm by recreating the painter instance again and again instead of just invalidating the draw (which is happening since the state is being read in a DrawScope).

compose/snippets/build.gradle.kts
```diff
@@ -46,6 +46,7 @@ android {
         }
 
         getByName("release") {
+            signingConfig = signingConfigs.getByName("debug")
```
bsagmoe, line 49: Was this just for some local testing with a release build for performance? Don't think it's a major issue if we keep this, but maybe add a comment noting why this is here.

bsagmoe (approved):

compose/snippets/build.gradle.kts
riggaroo: its to be able to make a release version, its not an issue as its just a sample app that wont be released to google play

PR #949, opened 2026-06-15
Add snippets for Grid named area

chikoski:
This pull request adds snippets describing Grid name area.  The functionality is currently available in SNAPSHOT.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/layouts/grid/NamedArea.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/layouts/grid/NamedArea.kt
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:OptIn(
+    ExperimentalGridApi::class, ExperimentalMediaQueryApi::class,
+    ExperimentalComposeUiApi::class
+)
+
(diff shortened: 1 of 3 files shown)


prateekbatra-g (approved):

gradle/libs.versions.toml
```diff
@@ -42,7 +42,7 @@ appcompat = "1.7.1"
 coil = "2.7.0"
 # @keep
 compileSdk = "37"
-compose-latest = "1.11.1"
+compose-latest = "1.12.0-SNAPSHOT"
```
kkuan2011: is this safe to change - as in are any other teams relying on compose-latest? and when will we update or remove the snapshot version?

compose/snippets/src/main/java/com/example/compose/snippets/layouts/grid/NamedArea.kt
```diff
@@ -0,0 +1,182 @@
+@file:OptIn(
```
kkuan2011, line 17: Missing copyright header. I think you can run './gradlew spotlessApply' to fix that and the build.

chikoski: Thanks for catching this. I misunderstood the copyright is automatically added on GitHub. Added as suggested.

gradle/libs.versions.toml
chikoski: I changed it to 1.12.0-beta01 as  it includes the named-area feature.

gradle/libs.versions.toml
```diff
@@ -42,7 +42,7 @@ appcompat = "1.7.1"
 coil = "2.7.0"
 # @keep
 compileSdk = "37"
-compose-latest = "1.11.1"
+compose-latest = "1.12.0-beta01"
```
kkuan2011, line 47: Could you check with Rebecca F. on whether this change is ok? I believe we wanted to keep compose on a stable version, and any code that needs to use alpha/beta version can stay on a branch (instead of being merged to main)

kkuan2011 (commented):
Rest of the change looks good, but I think the build is still failing?

gradle/libs.versions.toml
chikoski: Thanks for flagging this, and the feature used in the snippet is available from 1.12.0-beta01. This pull request will update the adaptive-layouts branch. I am remove the line when I create a pull request to get the adaptive-layout branch to the main.

PR #969, opened 2026-06-25
add snippets on how to request projected permissions from phone activity

trambui09:
Add snippets on how to request projected permissions from phone activity using the deviceId overload.


xr/src/main/java/com/example/xr/projected/PermissionsFromPhoneActivity.kt
```diff
@@ -0,0 +1,90 @@
...
+        const val REQUEST_CODE_GLASSES_CAMERA = 1001
+    }
+
+    // [START androidxr_projected_permissions_from_phone_activity_permission_result_callback]
+    override fun onRequestPermissionsResult(
+        requestCode: Int,
+        permissions: Array<out String>,
+        grantResults: IntArray,
+        deviceId: Int
+    ) {
```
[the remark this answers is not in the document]
trambui09: Done!

xr/src/main/java/com/example/xr/projected/PermissionsFromPhoneActivity.kt
```diff
@@ -0,0 +1,91 @@
...
+                // [START androidxr_projected_permissions_from_phone_activity_request_permission]
+                // Request the projected permission from phone activity
+                requestPermissions(
+                    arrayOf(Manifest.permission.CAMERA),
+                    REQUEST_CODE_GLASSES_CAMERA,
+                    projectedDeviceId
+                )
+                // [END androidxr_projected_permissions_from_phone_activity_request_permission]
+            }
+        } catch (e: IllegalStateException) {
```
devbridie, line 91: In particular: this is caused by [`@throws IllegalStateException if the projected device was not found`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:xr/projected/projected/src/main/kotlin/androidx/xr/projected/ProjectedContext.kt;l=66;drc=3273c0c7f6d72970b6db06835b99f4fe96355912), could we be specific here?

xr/src/main/java/com/example/xr/projected/PermissionsFromPhoneActivity.kt
```diff
@@ -0,0 +1,91 @@
...
+            // [END androidxr_projected_permissions_from_phone_activity_has_permission]
+
+            if (hasPermission) {
+                // Proceed with camera features
+            } else {
+                // [START androidxr_projected_permissions_from_phone_activity_request_permission]
+                // Request the projected permission from phone activity
+                requestPermissions(
+                    arrayOf(Manifest.permission.CAMERA),
+                    REQUEST_CODE_GLASSES_CAMERA,
```
devbridie, line 86: Please either add a comment about `REQUEST_CODE_GLASSES_CAMERA` or [let the system manage the code](https://developer.android.com/training/permissions/requesting#allow-system-manage-request-code); without context `REQUEST_CODE_GLASSES_CAMERA` looks like a constant that you might get from either Jetpack XR or some other library as opposed to a developer-defined constant.

xr/src/main/java/com/example/xr/projected/PermissionsFromPhoneActivity.kt
trambui09: Thanks for calling this out! Added!

xr/src/main/java/com/example/xr/projected/PermissionsFromPhoneActivity.kt
trambui09: I added a comment about REQUEST_CODE_GLASSES_CAMERA being a dev defined constant as it's not possible to let the system manage the request code when requesting for projected hardware permission due to the deviceId being required.

devbridie (approved):

PR #978, opened 2026-07-09
Jetpack XR library update for July '26

trambui09:
Jetpack XR library update for July '26

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,4 +26,4 @@ android.enableR8.fullMode=true
 
 # Use an AndroidX snapshot build.
 # https://androidx.dev/snapshots/builds
-# snapshotVersion=14793336
+ snapshotVersion=15770982
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,4 +26,4 @@ android.enableR8.fullMode=true
 
 # Use an AndroidX snapshot build.
 # https://androidx.dev/snapshots/builds
- snapshotVersion=15770982
+ snapshotVersion=15680104
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,4 +26,4 @@ android.enableR8.fullMode=true
 
 # Use an AndroidX snapshot build.
 # https://androidx.dev/snapshots/builds
- snapshotVersion=15680104
+# snapshotVersion=15680104
--- a/xr/src/main/java/com/example/xr/arcore/Anchors.kt
+++ b/xr/src/main/java/com/example/xr/arcore/Anchors.kt
@@ -23,7 +23,7 @@ import androidx.xr.runtime.Config
 import androidx.xr.runtime.Session
 import androidx.xr.runtime.SessionConfigureSuccess
 import androidx.xr.runtime.math.Pose
-import androidx.xr.scenecore.AnchorEntity
+import androidx.xr.scenecore.AnchorSpace
 import androidx.xr.scenecore.Entity
 import androidx.xr.scenecore.scene
 
@@ -59,7 +59,7 @@ private fun attachEntityToAnchor(
     anchor: Anchor
 ) {
     // [START androidxr_arcore_entity_tracks_anchor]
-    AnchorEntity.create(session, anchor).apply {
+    AnchorSpace.create(session, anchor).apply {
         parent = session.scene.activitySpace
         addChild(entity)
     }
--- a/xr/src/main/java/com/example/xr/arcore/Hands.kt
+++ b/xr/src/main/java/com/example/xr/arcore/Hands.kt
@@ -17,6 +17,7 @@
 package com.example.xr.arcore
 
 import android.app.Activity
+import android.content.Context
 import android.os.Bundle
 import android.util.Log
 import androidx.activity.ComponentActivity
@@ -151,10 +152,12 @@ class GenerateHandJointData : ComponentActivity() {
     @OptIn(FlowPreview::class)
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        val session = (Session.create(context = this) as SessionCreateSuccess).session
-        val config = Config.Builder(session.config).setHandTracking(HandTrackingMode.BOTH).build()
-        session.configure(config)
+
         lifecycleScope.launch {
+            val session = (Session.create(context = this as Context) as SessionCreateSuccess).session
+            val config = Config.Builder(session.config).setHandTracking(HandTrackingMode.BOTH).build()
+            session.configure(config)
+
             Hand.right(session).state.sample(500.milliseconds).collect { rightHandState ->
                 val bufferString = buildString {
                     append("mapOf(")
--- a/xr/src/main/java/com/example/xr/arcore/Hands.kt
+++ b/xr/src/main/java/com/example/xr/arcore/Hands.kt
@@ -17,7 +17,6 @@
 package com.example.xr.arcore
 
 import android.app.Activity
-import android.content.Context
 import android.os.Bundle
 import android.util.Log
 import androidx.activity.ComponentActivity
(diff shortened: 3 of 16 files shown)


xr/src/main/java/com/example/xr/arcore/Hands.kt
```diff
@@ -151,10 +152,12 @@ class GenerateHandJointData : ComponentActivity() {
     @OptIn(FlowPreview::class)
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        val session = (Session.create(context = this) as SessionCreateSuccess).session
-        val config = Config.Builder(session.config).setHandTracking(HandTrackingMode.BOTH).build()
-        session.configure(config)
+
         lifecycleScope.launch {
+            val session = (Session.create(context = this as Context) as SessionCreateSuccess).session
```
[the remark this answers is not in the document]
trambui09: Done!

xr/src/main/java/com/example/xr/compose/SceneCoreEntity.kt
```diff
@@ -58,8 +57,8 @@ fun SceneCoreEntityExample() {
                         size: IntVolumeSize
                     ) {
                         val extents = FloatSize2d(
-                            Meter.fromPixel(size.width.toFloat(), density).toM(),
-                            Meter.fromPixel(size.height.toFloat(), density).toM(),
+                            size.width.toFloat(),
+                            size.height.toFloat()
                         )
```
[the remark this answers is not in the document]
trambui09: The suggest didn't work due to the unresolved reference `pxToMeters`

gradle.properties
```diff
@@ -26,4 +26,4 @@ android.enableR8.fullMode=true
 
 # Use an AndroidX snapshot build.
 # https://androidx.dev/snapshots/builds
-# snapshotVersion=14793336
+ snapshotVersion=15770982
```
kkuan2011: I forget, if we set a snapshot version, does this have impact on other teams?

xr/src/main/java/com/example/xr/arcore/Anchors.kt
```diff
@@ -23,7 +23,7 @@ import androidx.xr.runtime.Config
 import androidx.xr.runtime.Session
 import androidx.xr.runtime.SessionConfigureSuccess
 import androidx.xr.runtime.math.Pose
-import androidx.xr.scenecore.AnchorEntity
```
vad710, line 26: Just to confirm that we are handling this name change properly - is this a breaking change or is `AnchorEntity` still available and marked as depracated?

xr/src/main/java/com/example/xr/arcore/Hands.kt
```diff
@@ -151,10 +151,12 @@ class GenerateHandJointData : ComponentActivity() {
     @OptIn(FlowPreview::class)
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        val session = (Session.create(context = this) as SessionCreateSuccess).session
-        val config = Config.Builder(session.config).setHandTracking(HandTrackingMode.BOTH).build()
-        session.configure(config)
+
         lifecycleScope.launch {
+            val session = (Session.create(context = this@GenerateHandJointData) as SessionCreateSuccess).session
```
vad710, line 156: Just for my own understanding, what does the syntax of `@GenerateHandJointData` do here?

xr/src/main/java/com/example/xr/compose/Orbiter.kt
```diff
@@ -59,8 +58,8 @@ private fun OrbiterExampleSubspace() {
             SubspaceModifier
                 .height(824.dp)
                 .width(1400.dp)
-                .transformingResizable()
```
vad710, line 62: Same question here - how is this deprecation being handled?

Do we have enough changes that we should think about doing a migration guide? Or perhaps a few things we should highlight in the Beta Blog post in August?

xr/src/main/java/com/example/xr/compose/SceneCoreEntity.kt
```diff
@@ -58,8 +57,8 @@ fun SceneCoreEntityExample() {
                         size: IntVolumeSize
                     ) {
                         val extents = FloatSize2d(
-                            Meter.fromPixel(size.width.toFloat(), density).toM(),
-                            Meter.fromPixel(size.height.toFloat(), density).toM(),
+                            size.width.toFloat(),
+                            size.height.toFloat()
```
vad710, line 61: is the data we are returning from these values fundamentally different now?

I am trying to understand if we have a sneaky breaking change that might be hard to detect and leading to bugs.

xr/src/main/java/com/example/xr/projected/ProjectedMainActivity.kt
```diff
@@ -158,15 +160,15 @@ fun HomeScreen(
...
+            ActionCard(
                 title = { Text("Permission Required") },
                 action = { Button(onClick = onClose) { Text("Exit") } }
             ) {
                 Text("Camera access is needed to use AI glasses features.")
                 Button(onClick = onRetryPermission) { Text("Retry") }
             }
         } else if (isVisualUiSupported) {
-            Card(
+            ActionCard(
```
vad710, line 171: This is an interesting API change. Same concerns re depracation.

vad710 (approved):
LGTM - left several comments/questions about how we are handling the breaking change and how we might best set developers for success to navigate these changes.

xr/src/main/java/com/example/xr/arcore/Hands.kt
yrezgui: `this` refers to `lifecycleScope` given that it's within its block. By putting `@GenerateHandJointData`, you go one level higher, which is an `Activity` and it's castable as a `Context

gradle.properties
yrezgui: If other teams were relying on an older snapshot release and the API changing between the publication of a snippet but that's extremely rare. Once the release is ready, we don't really change it, we upgrade the version number. I would say if we uncomment this line, it should be commented back within 2 weeks (Jetpack release cycle)

yrezgui (approved):

devbridie:
Moved the PR back to Draft status due to `snapshotVersion` change - that change is not meant to land in `main`.

gradle.properties
devbridie: The intention is that the PR author reverts the change to this file - this change is to help XR team members synchronize on the same snapshot version. When Jetpack goes live, we revert the change to `snapshotVersion` and merge into `main`.

xr/src/main/java/com/example/xr/arcore/Anchors.kt
trambui09: It's a breaking change. Any code using the old AnchorEntity class will need to be updated to use AnchorSpace.

xr/src/main/java/com/example/xr/compose/Orbiter.kt
trambui09: I think @amyZepp would be a good person to make that call. What do you think Amy?

xr/src/main/java/com/example/xr/projected/ProjectedMainActivity.kt
trambui09: If the glimmer Card composable has an action parameter, it is a breaking change (which it is in this case). The glimmer Card composable still exists.

xr/src/main/java/com/example/xr/arcore/Anchors.kt
devbridie: No breakage for downstream clients just yet - see [cs.android.com](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:xr/scenecore/scenecore/src/main/java/androidx/xr/scenecore/AnchorEntity.kt;l=57;drc=c30c2b1e983b61cb3ffe4ab5da01d908e2b10e71); `AnchorEntity` is being marked `@Deprecated`.

kkuan2011:
Regarding the snapshot version, sounds good to keep this change on a branch, and not merge it into main.

PR #980, opened 2026-07-10
Update EngageWorker snippet to use V2 isServiceAvailable API

amulyagaur:
Updates the EngageWorker.kt snippet to use the new V2 version of isServiceAvailable API supporting ClusterType filtering.


huewu:
Hi @amulyagaur, 

Shouldn't we update the dependency version, as well? (1.5.2 > 1.6.0). Every code in the snippet should be compilable without error and warning. So, please double check it. 

Thanks.

amulyagaur (commented):
Update Engage SDK Integration Skill snippets to use V2 isServiceAvailable API.

amulyagaur:
> Hi @amulyagaur,
> 
> Shouldn't we update the dependency version, as well? (1.5.2 > 1.6.0). Every code in the snippet should be compilable without error and warning. So, please double check it.
> 
> Thanks.

Hi, Thanks for pointing this out. I have made the version changes.

shashvatgupta-49 (approved):

huewu (approved):
LGTM.

amulyagaur:
cc @kkuan2011

yrezgui (approved):

PR #981, opened 2026-07-11
Fix formatting issue in coroutines snippet

kkuan2011:
Minor nit, but adding the /** **/ back to a function for the first snippet on this page:
https://developer.android.com/kotlin/coroutines/coroutines-adv


askandroid (approved):
looks good

PR #990, opened 2026-07-23
[TV] Compose browse snippets

pflammertsma:


tv/build.gradle.kts
```diff
@@ -61,6 +61,7 @@ dependencies {
     implementation(libs.androidx.compose.ui.tooling.preview)
     implementation(libs.androidx.compose.material3)
     implementation(libs.tv.compose.material)
+    implementation(libs.tv.foundation)
```
chikoski: This is trivial, but proposing to remove tv-foundation from the dependencies.

chikoski (approved):
LGTM. I left a  minor comment.

tv/build.gradle.kts
pflammertsma: Good call, I don't know why I didn't think of that!

pflammertsma: Done.

PR #994, opened 2026-07-23
[TV] Add Leanback-to-Compose TV migration snippets

pflammertsma:


tv/src/main/java/com/example/tv/playback/ImmersiveBrowseScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+                        horizontalArrangement = Arrangement.End
+                    ) {
+                        Button(
+                            onClick = { /* Search */ },
+                            modifier = Modifier.focusRequester(topBarFocusRequester)
+                        ) { Text("Search") }
+                    }
+                }
+
+                categories.entries.forEachIndexed { catIndex, (categoryName, videos) ->
```
chikoski: [itemsIndexed](https://developer.android.com/reference/kotlin/androidx/compose/foundation/lazy/LazyListScope#(androidx.compose.foundation.lazy.LazyListScope).itemsIndexed(kotlin.Array,kotlin.Function2,kotlin.Function2,kotlin.Function3)) could simplify the code as follows;

```kotlin
itemsIndex(categories) { index, (categoryName, videos) ->
// snip
}
```

tv/src/main/java/com/example/tv/playback/Media3PlaybackScreen.kt
```diff
@@ -0,0 +1,175 @@
...
+// [START android_compose_tv_media3_playback]
+@OptIn(UnstableApi::class)
+@Composable
+fun Media3PlaybackScreen(
+    video: Video,
+    onFinish: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    val context = LocalContext.current
+    val exoPlayer = remember { ExoPlayer.Builder(context).build() }
```
chikoski: Proposing to update the `exoPlayer` value when active context is updated:

```kotlin
val exoPlayer = remember(context) { 
// snip;
}

File: tv/src/main/java/com/example/tv/playback/Media3PlaybackScreen.kt [hunk 3]
```diff
@@ -0,0 +1,175 @@
...
+            delay(5000)
+            showControls = false
+        }
+    }
+
+    LaunchedEffect(video) {
+        focusRequester.requestFocus()
+        scheduleAutoHide()
+
+        val listener = object : Player.Listener {
```
chikoski, line 84: I am wondering if the ExoPlayer object adds new listener every time the user chooses new video to play. I'd suggest to add DisposableEffect and move the code creating and setting lister object outside of the LaunchedEffect to it.

chikoski (commented):
I quickly reviewed the snippets and left some comments. Please take a look. 

Overall, the snippets seems slightly long, which might make it difficult for developers to digest.  To make it easier to read, I'd suggest to focus on the essential part, such as how to coordinate list item selection and background image update for the immersive list snippet.

Key event handling and updating the visibility of playback controls in the video player screen might be a bit out of scope here, or could be omitted to keep the focus clear.

pflammertsma:
Noted that the snippets are long, but they aren't really shared with developers; they're used for the migration skill. The skill is hosted on DAC, so we must have all snippets on GitHub.

tv/src/main/java/com/example/tv/playback/ImmersiveBrowseScreen.kt
pflammertsma: Resolved throughout.

tv/src/main/java/com/example/tv/playback/Media3PlaybackScreen.kt
pflammertsma: Done.

chikoski (approved):

PR #1004, opened 2026-08-03
Add Navigation 3 deep link snippets

bsagmoe:
This PR adds code snippets for Navigation 3 deep links documentation.


bsagmoe:
Note: The build will fail until the dependency is updated to the non-snapshot beta release tomorrow, which includes the RequestExtras changes.

datastore/build.gradle.kts
```diff
@@ -24,7 +24,7 @@ android {
 
     defaultConfig {
         applicationId = "com.example.datastore.snippets"
-        minSdk = 23
+        minSdk = 24
```
yrezgui, line 27: Why upgrading the minSdk of this module?

yrezgui (dismissed):

datastore/build.gradle.kts
bsagmoe: The build was failing otherwise (might be related to the fact that I'm using a SNAPSHOT build for Nav 3). The beta for Nav 3 1.2 comes out tomorrow, so I'll double check if this change is necessary then.

bsagmoe: It's necessary because the `datastore` module depends on Nav 3 and, with the recent Jetpack shift to minSdk 24 (https://issuetracker.google.com/474169350), Nav 3 1.2.0 has a minSdk of 24.

dturner (approved):

PR #1013, opened 2026-08-07
Adding input example project

acianrice:
As requested by Adrian, this is a small example projected of using different inputs on the glasses (back gesture, generic motion inputs, and the camera action). Not sure if this is exactly what folks were looking for.


trambui09:
Hi @acianrice, since this is your first time contributing, please make sure to sign the google CLA.

This PR contains a lot of boilerplate code. I would suggest distilling it down to the minimum code needed for the snippets. See https://github.com/android/snippets/tree/main/xr/src/main/java/com/example/xr for some examples. This will help with maintenance.

PR #1024, opened 2026-08-13
Update HealthConnectManager.kt, specifically the "!=null" 

brenda-shaw:
Requesting to update "!=null". The standalone APK version of Health Connect on Android 13/12, or across specific IPC serialization boundaries—ReadRecordsResponse.pageToken can return an empty string ("") instead of null when pagination is complete.

Because "" != null evaluates to true, the loop condition fails to terminate. This causes severe downstream effects for integrating apps, including infinite loops, rate limit exhaustion, and potential application crashes.


brenda-shaw:
Can someone please approve this? There's only one line that's being updated, thanks

lindazhao-google (approved):

kkuan2011 (approved):

PR #1030, opened 2026-08-13
Updating snippet to use coil3 artifact

wardlauren:
Updating snippet to use coil3 artifact.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/images/LoadingImagesSnippets.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/images/LoadingImagesSnippets.kt
@@ -21,7 +21,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
-import coil.compose.AsyncImage
+import coil3.compose.AsyncImage
 import com.bumptech.glide.integration.compose.ExperimentalGlideComposeApi
 import com.bumptech.glide.integration.compose.GlideImage
 import com.example.compose.snippets.R
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -41,7 +41,10 @@ androidx-work-runtime = "2.11.2"
 androidxHiltNavigationCompose = "1.4.0"
 appcompat = "1.7.1"
 camera-camera2 = "1.6.1"
+# Coil 2
 coil = "2.7.0"
+# Coil 3
+coil3 = "3.5.0"
 # @keep
 compileSdk = "37"
 compose-latest = "1.12.0"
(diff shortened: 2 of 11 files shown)


nahidhossain199612-lgtm (approved):

riggaroo (approved):

yrezgui (changes requested):
I need to check why we have both of these dependencies

tv/build.gradle.kts
```diff
@@ -64,6 +64,8 @@ dependencies {
     implementation(libs.androidx.fragment.ktx)
     implementation(libs.androidx.work.runtime.ktx)
     implementation(libs.coil.kt.compose)
+    implementation(libs.coil3.compose)
```
yrezgui: Why are we adding these libraries in this module? The imports are still referecing coil2

pflammertsma: I'm a bit confused why we're adding coil3 as a dependency without changing the import; [see the usages](https://github.com/search?q=repo%3Aandroid%2Fsnippets+path%3Atv%2Fsrc%2Fmain%2Fjava%2Fcom%2Fexample%2Ftv+AsyncImage&type=code).

pflammertsma: Please either remove the implementation or add the coil3 import.

pflammertsma (changes requested):

compose/snippets/src/main/java/com/example/compose/snippets/images/LoadingImagesSnippets.kt
```diff
@@ -21,7 +21,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
-import coil.compose.AsyncImage
+import coil3.compose.AsyncImage
```
yrezgui, line 24: Why aren't we updating all the imports in the compose module to reference coil3?

yrezgui (changes requested):

compose/snippets/src/main/java/com/example/compose/snippets/images/LoadingImagesSnippets.kt
wardlauren: I believe I've updated all the imports in the Compose module now to reference coil3.

tv/build.gradle.kts
wardlauren: Removed the library from this module.

wardlauren: Removed the library from this module.

gradle/libs.versions.toml
```diff
@@ -41,7 +41,10 @@ androidx-work-runtime = "2.11.2"
 androidxHiltNavigationCompose = "1.4.0"
 appcompat = "1.7.1"
 camera-camera2 = "1.6.1"
+# Coil 2
 coil = "2.7.0"
```
riggaroo, line 45: might want to remove this

PR #1031, opened 2026-08-14
Migrate Compose Glance and Navigation snippets to GitHub

hamen:
Migrate Compose Glance and Navigation snippets to GitHub.

Migrate hardcoded code snippets across Compose Glance, and Edge-to-edge documentation pages to GitHub-hosted code snippets in the android/snippets repository.

Affected Pages:

* [Build UI with Glance](https://developer.android.com/develop/ui/compose/glance/build-ui)
* [Handle errors with Glance](https://developer.android.com/develop/ui/compose/glance/error-handling)
* [Set up Edge-to-edge](https://developer.android.com/develop/ui/compose/system/setup-e2e)
* [Add support for the predictive back gesture](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture)
* [Add support for predictive back animations](https://developer.android.com/guide/navigation/custom-back/support-animations)


compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
```diff
@@ -1082,3 +1092,113 @@ class SyncService : Service() {
...
+            rv.setOnClickPendingIntent(R.id.error_icon, getErrorIntent(context, throwable))
+            AppWidgetManager.getInstance(context).updateAppWidget(appWidgetId, rv)
+        }
+        // [END android_compose_glance_on_composition_error_override]
+
+        // [START android_compose_glance_error_get_intent]
+        private fun getErrorIntent(context: Context, throwable: Throwable): PendingIntent {
+            val intent = Intent(context, UpgradeToHelloWorldPro::class.java)
+            intent.action = "widgetError"
+            // [START_EXCLUDE]
```
kkuan2011: Can you add silent to the tag so ... doesn't show up in the visible snippet on DAC?

// [START_EXCLUDE silent]

compose/snippets/src/main/java/com/example/compose/snippets/predictiveback/PredictiveBackSnippets.kt
```diff
@@ -200,3 +208,58 @@ private enum class DrawerState {
...
+                animate(xOffset, 0f) { value, _ -> xOffset = value }
+            }
+        }
+    }
+    Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
+        Surface(
+            modifier = Modifier
+                .size(200.dp)
+                .scale(scale)
+                .offset { IntOffset(x = xOffset.dp.roundToPx(), y = 0) },
```
kkuan2011, line 259: Is this the preferred way to do the offset? compared to what was in the original code?

.offset(x = xOffset.dp, y = 0.dp)

hamen: Yes, I followed the suggestion AS and the linter gave me.

compose/snippets/src/main/java/com/example/compose/snippets/glance/GlanceSnippets.kt
hamen: Done ✅

kkuan2011 (approved):

compose/snippets/src/main/java/com/example/compose/snippets/system/EdgeToEdgeSnippets.kt
```diff
@@ -236,3 +241,45 @@ fun DialogExample(){
     ) { /* Content */ }
     // [END android_compose_e2e_dialog]
 }
+
+private class SetupE2ESnippetsActivity : ComponentActivity() {
+
+    // [START android_compose_e2e_enable]
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        enableEdgeToEdge()
```
alabiaga, line 250: Now that we are migrating these to snippets, it is more useful as the developer can go to this file and see that enableEdgeToEdge is an extension function. The [documentation](https://developer.android.com/develop/ui/compose/system/setup-e2e#enable-edge-to-edge) needs to be modified as the enableEdgeToEdge link being referred to prior to showing the code snippet links to the method defined in WindowCompat vs the extension function defined in package androidx.activity.ComponentActivity that is being used in the snippet.

compose/snippets/src/main/java/com/example/compose/snippets/system/EdgeToEdgeSnippets.kt
```diff
@@ -236,3 +241,45 @@ fun DialogExample(){
...
+        // ...
+    }
+    // [END android_compose_e2e_enable]
+
+    private fun immersiveModeExample(window: Window) {
+        // [START android_compose_e2e_immersive_mode]
+        val windowInsetsController = WindowCompat.getInsetsController(window, window.decorView)
+
+        // Hide the system bars.
```
alabiaga, line 259: nitpick and I wish others would have said that same in my changes but we should be consistent with whether we omit or end in a period for these comments.

I think we should always end with a period, ".".

alabiaga (commented):
initial pass, will look again.

compose/snippets/src/main/java/com/example/compose/snippets/system/EdgeToEdgeSnippets.kt
hamen: Hey Aaron, things are moving fast on these pages. Looks like that one has
already been updated: the enableEdgeToEdge link in that section now points at
the ComponentActivity extension, the androidx.activity one.

On our side the snippet imports androidx.activity.enableEdgeToEdge and calls it
bare inside a ComponentActivity, so it is the extension. The WindowCompat
import in that file is for the immersive mode and status bar icon snippets, not
for this one.

So I think we're already where you wanted this to land.

compose/snippets/src/main/java/com/example/compose/snippets/system/EdgeToEdgeSnippets.kt
hamen: Agreed, let's make it consistent. Added the period to the comments this PR
brings in, eight across the two files.

Two of them are copied straight from the setup-e2e page, which publishes them
without one, so the page picks up the change through the include.

Left the TODO in GlanceSnippets.kt alone, "// TODO: handle error." reads odd
with a full stop. Happy to add it if you want the rule applied everywhere.

compose/snippets/src/main/java/com/example/compose/snippets/system/EdgeToEdgeSnippets.kt
alabiaga: Right, i pushed that change. Thanks Ivan

alabiaga (approved):

MagicalMeghan (approved):

PR #1041, opened 2026-08-20
Add Compose Grid snippets for overlay and tabletop layouts

prateekbatra-g:
- Grid Overlays: Implements GridOverlayCard with overlapping named area spans (Areas, gridConfig) and a TextCard helper.

- Tabletop Adaptive Layout: Implements GridTabletop using mediaQuery and Posture.Tabletop to split controls across the fold.

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/layouts/grid/GridByExamples.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/layouts/grid/GridByExamples.kt
@@ -22,18 +22,25 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.ExperimentalGridApi
 import androidx.compose.foundation.layout.Grid
+import androidx.compose.foundation.layout.GridConfigurationScope
 import androidx.compose.foundation.layout.GridTrackSize
 import androidx.compose.foundation.layout.fillMaxHeight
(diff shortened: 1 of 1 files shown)


chikoski (commented):
Overall, it looks good to me! I left some comments, pease take a look.

compose/snippets/src/main/java/com/example/compose/snippets/layouts/grid/GridByExamples.kt
```diff
@@ -209,4 +219,119 @@ private fun WideClassicCardPreview() {
...
+        Text(text = text, color = Color.White, fontSize = 24.sp)
+    }
+}
+// [END android_compose_grid_overlay_helper_text_card]
+
+// [START android_compose_grid_tabletop_config_setup]
+enum class VideoPlayerArea { Player, Controller }
+
+// Default: Single area layout with overlay (controls on top of player)
+val defaultConfig: GridConfigurationScope.() -> Unit = {
```
chikoski: I'd suggest making `defaultConfig` and `tabletopConfig` as a members of a custom object, such as `PlayerGridConfig`. It would clarify our intention, i.e. the component switches these two configurations dynamically. It could also reduce the number of globally exposed symbols.

prateekbatra-g: added PTAL

chikoski (approved):

PR #1049, opened 2026-08-25
Add Kotlin Parcelize snippets

erikrodriguez-se:
# Snippets migration - Kotlin parcelize

Snippets for the Parcelize guide at [developer.android.com/kotlin/parcelize](https://developer.android.com/kotlin/parcelize), extracted into `kotlin/src/main/kotlin/com/example/android/basics/` (`Parcelize.kt` and `ParcelizeDataClass.kt`) with region tags, plus the parcelize plugin in `kotlin/build.gradle.kts`. The multiplatform snippets are added to the `kmp/shared` module (package `com.example.kmp.snippets`).

Fifteen of the page's 24 code blocks are extracted. Most are verbatim; the differences that exist fall into three kinds, described below.

| Region tag (`android_kotlin_parcelize_*`) | Page section | Verbatim | Deviation | Rendered / page |
|---|---|---|---|---|
| `_basic` | intro | no | import shown as a comment; colon spacing | 4 / 4 |
| `_companion_parceler` | intro | yes | — | 12 / 12 |
| `_external_class_parceler` | Custom Parcelers | yes | — | 9 / 9 |
| `_type_parceler_class` | Custom Parcelers | yes | — | 4 / 4 |
| `_type_parceler_property` | Custom Parcelers | yes | — | 3 / 3 |
| `_type_parceler_type` | Custom Parcelers | yes | — | 3 / 3 |
| `_parcelable_creator` | Create data from Parcel | no | import shown as a comment | 5 / 5 |
| `_ignored_on_parcel` | Skip properties from serialization | no | colon spacing | 10 / 10 |
| `_raw_value` | `Parcel.writeValue` | no | colon spacing | 2 / 2 |
| `_sealed_class` | Sealed classes and interfaces | no | colon spacing ×5 | 8 / 8 |
| `_data_class` | Experimental / Data class serializer | no | `@file:OptIn` shown as a comment | 6 / 6 |
| `_data_class_wrapper` | Experimental / Data class serializer | no | colon spacing | 8 / 8 |
| `_inheritance_base` | Experimental / Non `val` or `var` params | no | colon spacing ×2 | 11 / 11 |
| `_multiplatform_common` | Setup Parcelize for Kotlin multiplatform | no | package shown as a comment; colon spacing | 19 / 19 |
| `_multiplatform_platform` | Setup Parcelize for Kotlin multiplatform | no | package shown as a comment | 6 / 6 |

### The three kinds of deviation

**Colon spacing.** The page writes `): Parcelable`; this repo's ktlint config rewrites it to `) : Parcelable`. This is unavoidable — `spotlessApply` makes the change when run — and it accounts for most of the non-verbatim rows above.

**File-top declarations shown as a comment.** The `_basic` and `_parcelable_creator` blocks open with an `import`, `_data_class` opens with `@file:OptIn(kotlinx.parcelize.Experimental::class)`, and the two `_multiplatform_` blocks open with a `package` declaration. Kotlin requires those at the top of the file, above the `package` declaration and outside any snippet region. To keep them visible in the published snippet, the real line lives at the top of the file (uncommented, so it compiles) and a **commented copy** is placed inside the region. The published snippet therefore shows, for example, `// import kotlinx.parcelize.Parcelize` where the page shows a live `import` statement. The page's blocks would be updated to match.

**One experimental block omitted.** The page's "Non `val` or `var` parameters" section has two code blocks: `_inheritance_base` (the non-experimental approach, extracted) and an experimental variant that requires `experimentalCodeGeneration=true`. That flag is a parcelize compiler-plugin argument with module-level granularity — it cannot be scoped to a single snippet, and enabling it module-wide would compile the `kotlin` module's unrelated snippets (coroutines, flow testing) under experimental codegen. The experimental block is therefore not extracted, and the flag is not set. `_inheritance_base` alone still demonstrates the section's point.

### Multiplatform snippets

The Kotlin 2.0+ multiplatform snippets are added to `kmp/shared`: `_multiplatform_common` in `commonMain` and `_multiplatform_platform` in `androidMain`. Two notes:

- **iOS actuals.** The common code declares `expect interface MyParcelable` and `expect annotation class MyIgnoredOnParcel`, so Kotlin requires an `actual` in every target — including iOS. The page says these "can be empty" but does not show them, so `iosMain` gets a small file with empty actuals. It is not a page snippet, so it carries no region tags.
- **Gradle config.** The `additionalAnnotation` compiler argument is applied in `kmp/shared/build.gradle.kts` but left untagged, matching how the page publishes build config (hand-maintained). The page's `example` package becomes `com.example.kmp.snippets`, so the `additionalAnnotation` value points at `com.example.kmp.snippets.MyParcelize`.

### Code not extracted

- **Gradle config** — the [Groovy](https://developer.android.com/kotlin/parcelize#groovy) `plugins` block (this repo is Kotlin DSL only), the [Kotlin DSL](https://developer.android.com/kotlin/parcelize#kts) `plugins` block, and the `experimentalCodeGeneration` block. Build-config blocks are published in their portable `id("…")` form on the page rather than this repo's version-catalog form, so they are left hand-maintained. (The `additionalAnnotation` block is applied but untagged — see "Multiplatform snippets" above.)
- **Java** — the `UserCreator` block under [Create data from Parcel](https://developer.android.com/kotlin/parcelize#create_data_from_parcel). Kotlin only.
- **Multiplatform, pre-Kotlin 2.0** — the two `expect`/`actual` aliasing blocks under [Setup Parcelize for Kotlin multiplatform](https://developer.android.com/kotlin/parcelize#setup_parcelize_for_kotlin_multiplatform). That approach is unsupported in Kotlin 2.0 and higher, which this project uses, so it will not compile. The Kotlin 2.0+ blocks from the same section are extracted — see "Multiplatform snippets" above.
- **The experimental inheritance block** and **the `// ERROR: not allowed` example**, both under [Non val or var parameters in primary constructor](https://developer.android.com/kotlin/parcelize#non_val_or_var_parameters_in_primary_constructor). See "One experimental block omitted" above; the ERROR example cannot compile by design.

### Deprecations

None. Ten imports, all current — `android.os.Parcel`, `android.os.Parcelable` and eight `kotlinx.parcelize` symbols. A forced full recompile produced zero warnings.

### DIFF
--- a/kotlin/build.gradle.kts
+++ b/kotlin/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     alias(libs.plugins.android.application)
+    alias(libs.plugins.kotlin.parcelize)
 }
 
 kotlin {
@@ -26,6 +27,7 @@ android {
 
         getByName("release") {
             isMinifyEnabled = true
+            isShrinkResources = true
             proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"),
                 "proguard-rules.pro")
         }
--- a/kmp/shared/build.gradle.kts
+++ b/kmp/shared/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
     alias(libs.plugins.kotlin.multiplatform)
     alias(libs.plugins.android.kotlin.multiplatform.library)
+    alias(libs.plugins.kotlin.parcelize)
     alias(libs.plugins.android.lint)
 }
 
@@ -12,6 +13,10 @@ kotlin {
         namespace = "com.example.kmp.snippets.shared"
         compileSdk = libs.versions.compileSdk.get().toInt()
         minSdk = libs.versions.minSdk.get().toInt()
+
+        compilerOptions {
+            freeCompilerArgs.addAll("-P", "plugin:org.jetbrains.kotlin.parcelize:additionalAnnotation=com.example.kmp.snippets.MyParcelize")
+        }
 
         withHostTestBuilder {
         }
--- a/kmp/shared/src/androidMain/kotlin/com/example/kmp/snippets/Parcelize.kt
+++ b/kmp/shared/src/androidMain/kotlin/com/example/kmp/snippets/Parcelize.kt
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.kmp.snippets
+
+// [START android_kotlin_parcelize_multiplatform_platform]
+// Platform code
+// package example
+
+// No typealias for MyParcelize here
+actual typealias MyParcelable = android.os.Parcelable
+actual typealias MyIgnoredOnParcel = kotlinx.parcelize.IgnoredOnParcel
+// [END android_kotlin_parcelize_multiplatform_platform]
--- a/kmp/shared/src/commonMain/kotlin/com/example/kmp/snippets/Parcelize.kt
+++ b/kmp/shared/src/commonMain/kotlin/com/example/kmp/snippets/Parcelize.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.kmp.snippets
+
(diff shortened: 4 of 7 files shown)


kotlin/build.gradle.kts
```diff
@@ -26,6 +27,7 @@ android {
 
         getByName("release") {
             isMinifyEnabled = true
+            isShrinkResources = true
```
erikrodriguez-se, line 30: Not in the original fork PR but this is a linter nit worth addressing since we're touching the file

kkuan2011: sounds good, thanks for addressing that!

francescoo22:
> Multiplatform — all four expect/actual blocks under [Setup Parcelize for Kotlin multiplatform](https://developer.android.com/kotlin/parcelize#setup_parcelize_for_kotlin_multiplatform). Will not compile in a plain Android module.

Wouldn't it be possible to put Parcelize snippets in a multiplatform module so that we can add them as well? Since we are currently working on improving support for KMP projects using Parcelize, it would be nice to include snippets showing multiplatform usage if that is not too complicated. (I saw that there's already a multiplatform module in the project)

erikrodriguez-se:
> > Multiplatform — all four expect/actual blocks under [Setup Parcelize for Kotlin multiplatform](https://developer.android.com/kotlin/parcelize#setup_parcelize_for_kotlin_multiplatform). Will not compile in a plain Android module.
> 
> Wouldn't it be possible to put Parcelize snippets in a multiplatform module so that we can add them as well? Since we are currently working on improving support for KMP projects using Parcelize, it would be nice to include snippets showing multiplatform usage if that is not too complicated. (I saw that there's already a multiplatform module in the project)

@francescoo22 I added the two migrate-able snippets into the KMP module as requested and updated the PR description. Thanks!

francescoo22:
Thank you, looks good to me!

kkuan2011 (approved):

kkuan2011 (commented):
Looks good, but it looks like the iOS build is failing? Would you know why?

kkuan2011 (changes requested):

erikrodriguez-se:
> Looks good, but it looks like the iOS build is failing? Would you know why?

Sorry, this built cleanly locally for me and I didn't see the CI fail. It looks like the build script is asking for an older version of Xcode that the runner doesn't have. I will see if I can update it.

kkuan2011 (approved):

PR #1062, opened 2026-09-01
Add example of using FLAG_SECURE

alabiaga:
to be used for Android 17 documentation


robclifford27:
For consistency can you move the SecureWindowActivity.kt into a subfolder so it's clearly distinct from the intent snippet? Otherwise this looks fine to me

alabiaga:
> For consistency can you move the SecureWindowActivity.kt into a subfolder so it's clearly distinct from the intent snippet? Otherwise this looks fine to me

done. Thanks

robclifford27 (approved):

PR #1077, opened 2026-09-08
Nest CameraX snippets under :camera :camerax

barbaralaw:
## Summary
- Renames the existing `:camerax` Gradle module to `:camera` and nests camerax files as a sub-folder package within it so camera APIs can share a parent module instead of living at the repo root.
- Existing region tags within camerax files renamed to specify new package path
## Test plan
- [x] `./gradlew :camera:camerax:compileDebugKotlin`
- [x] `./gradlew :camera:camerax:build`
- [ ] Confirm CI `Build All` still succeeds on this path
- [x] Confirm no remaining references to the old `:camerax` module path


yrezgui:
Thanks for the pull request but we're not looking at expanding the camera snippets beyond camerax at the moment. I'm closing it for now

yrezgui:
I should look more at my emails (didn't know the context), sorry for the confusion @barbaralaw 🙏

settings.gradle.kts
```diff
@@ -40,7 +40,7 @@ include(
     ":kotlin",
     ":compose:snippets",
     ":datastore",
-    ":camerax",
+    ":camera:camerax",
```
kkuan2011: Thanks Barbara! Couple comments:

1. Can this just be called "camera"? I don't think we need to call out camerax directly 

2. Instead of directly moving the build + proguard files over to the camerax subdirectory, can we keep them at the root of the camera directory? so we can build the camera module directly 

3. Would you have the CL for the updates to the skill pages too? so we can submit the PR + CL close together.

barbaralaw: 1. & 2.  Yes, I've pushed an update to make camera the module and have camerax as a sub-folder package within that.
3. I will have that CL up asap today

camera/src/main/java/com/example/camera/snippets/camerax/Camera1ToCameraXSnippets.kt
```diff
@@ -49,7 +49,7 @@ import java.util.concurrent.Executor
 
 @Composable
 fun InitializeCameraSnippet() {
-  // [START android_camerax_initialize_provider]
+  // [START android_camera_camerax_initialize_provider]
```
kkuan2011: for simplicity and conciseness, we can leave the region tags as they were (so we don't duplicate the word camera and we minimize the diff here too).

Can you revert these files and then update the package name?

camera/build.gradle.kts
kkuan2011: is this import needed? I haven't seen it in the other module's build files

kkuan2011 (commented):
Looks good after comments addressed

camera/src/main/java/com/example/camera/snippets/camerax/Camera1ToCameraXSnippets.kt
barbaralaw: Thanks, I’ll restore the original `android_camerax_*` region tags so we don’t duplicate “camera” and the DAC includes stay valid. Package rename should stay as `com.example.camera.snippets.camerax` to match the new module layout.

camera/build.gradle.kts
barbaralaw: Good call, I'll drop the `JvmTarget` import and switch to `jvmToolchain(17)` like other modules.

kkuan2011 (approved):

madebymozart (approved):
LGTM

ashnohe (approved):

PR #1079, opened 2026-09-10
Migrate Display time-sensitive notifications snippets

hamen:
Code snippets are for:
- https://developer.android.com/develop/ui/views/notifications/time-sensitive

List of modifications:
- Time-sensitive page: Migrated 5 Kotlin code snippets to TimeSensitive.kt.
- Time-sensitive page: Added baseline_auto_awesome_24.xml vector drawable resource to resolve notification icon reference.
- Time-sensitive page: Preserved snippet lines with standard exclude tags for compilable state declaration.
- Time-sensitive page: Renamed the `Application` subclass in the channel snippet from `DACapp` to `MyApplication`. The class sits inside the region tag, so the page text needs the same rename.

Snippets not migrated:
- Line 24 because it is an AndroidManifest.xml permission block.
- Line 211 because it is a Java snippet cataloged for removal per Kotlin-first migration guidelines (D15).

### DIFF
--- a/compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
+++ b/compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.compose.snippets.notifications
+
+import android.Manifest
+import android.annotation.SuppressLint
+import android.app.Application
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.Service
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import android.os.IBinder
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.rememberLauncherForActivityResult
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.annotation.RequiresApi
+import androidx.compose.material3.Button
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.core.app.NotificationCompat
+import com.example.compose.snippets.R
+
+@RequiresApi(Build.VERSION_CODES.O)
+// [START android_notifications_time_sensitive_channel]
+class DACapp : Application() {
+    override fun onCreate() {
+        super.onCreate()
+        val channel = NotificationChannel(
+            CHANNEL_ID,
+            "High priority notifications",
+            NotificationManager.IMPORTANCE_HIGH
+        )
+
+        val notificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+        notificationManager.createNotificationChannel(channel)
+    }
+}
+// [END android_notifications_time_sensitive_channel]
+
+@Composable
+private fun NotificationPermissionSample() {
+    var hasNotificationPermission by remember { mutableStateOf(false) }
+    // [START android_notifications_time_sensitive_permission]
+    val permissionLauncher = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.RequestPermission(),
+        onResult = { hasNotificationPermission = it }
+    )
+    // [START_EXCLUDE]
+    val fillerCode = Unit
+    // [END_EXCLUDE]
+    Button(
+        onClick = {
+            if (!hasNotificationPermission) {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                    permissionLauncher.launch(Manifest.permission.POST_NOTIFICATIONS)
+                }
+            }
+        },
+    ) {
+        Text(text = "Request permission")
+    }
+    // [END android_notifications_time_sensitive_permission]
+}
+
+private class TimeSensitiveNotificationActivity : ComponentActivity() {
+
+    @SuppressLint("MissingPermission")
+    // [START android_notifications_time_sensitive_create]
+    private fun showNotification() {
+        val notificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+
+        val notificationBuilder =
+            NotificationCompat.Builder(this, CHANNEL_ID)
+                .setSmallIcon(R.drawable.baseline_auto_awesome_24)
+                .setContentTitle("HIGH PRIORITY")
+                .setContentText("Check this dog puppy video NOW!")
+                .setPriority(NotificationCompat.PRIORITY_HIGH)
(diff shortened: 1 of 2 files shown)


compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
```diff
@@ -0,0 +1,126 @@
...
+
+@Composable
+private fun NotificationPermissionSample() {
+    var hasNotificationPermission by remember { mutableStateOf(false) }
+    // [START android_notifications_time_sensitive_permission]
+    val permissionLauncher = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.RequestPermission(),
+        onResult = { hasNotificationPermission = it }
+    )
+    // [START_EXCLUDE]
```
kkuan2011, line 68: To simplify, you could replace lines 68-70 with

// ...

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
```diff
@@ -0,0 +1,126 @@
...
+import androidx.core.app.NotificationCompat
+import com.example.compose.snippets.R
+
+@RequiresApi(Build.VERSION_CODES.O)
+// [START android_notifications_time_sensitive_channel]
+class DACapp : Application() {
+    override fun onCreate() {
+        super.onCreate()
+        val channel = NotificationChannel(
+            CHANNEL_ID,
```
kkuan2011, line 49: Declare CHANNEL_ID explicitly in this file?

I think it's finding the declaration from NotificationsSnippets.kt (in the same package com.example.compose.snippets.notifications) which declares a non-private top-level val CHANNEL_ID = "channelId"

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
```diff
@@ -0,0 +1,126 @@
...
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.core.app.NotificationCompat
+import com.example.compose.snippets.R
+
+@RequiresApi(Build.VERSION_CODES.O)
```
kkuan2011, line 43: Remove this line and the import?

kkuan2011 (approved):
Once comments are addressed, you can add Aaron for review and remove from "draft" state, thanks!

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
hamen: Done in 101e393. Lines 68-70 now collapse into a single `// ...` placeholder inside the region tag.

No `[START_EXCLUDE]` block and no filler statement are necessary, because the lines around the placeholder compile on their own.

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
hamen: Done in 101e393. `TimeSensitive.kt` now declares `private const val CHANNEL_ID = "channelId"` at the top of the file, outside the region tags.

Your diagnosis was correct. The identifier resolved to the non-private top-level `val CHANNEL_ID` in `NotificationsSnippets.kt`. After I added the local declaration, the two files collided at package scope and the compiler reported an overload resolution ambiguity. I therefore also changed the declaration in `NotificationsSnippets.kt` to `private const val`, which keeps each file self-contained.

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
hamen: Done in 101e393. I removed `@RequiresApi(Build.VERSION_CODES.O)` and the `androidx.annotation.RequiresApi` import.

The annotation was redundant, because `minSdk` is 36 in `gradle/libs.versions.toml`, which is much higher than API level 26.

kkuan2011 (changes requested):
Ah actually looks like the build fails:

/home/runner/work/snippets/snippets/compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt:120: Error: To call Service.startForeground(), the <service> element of manifest file must have the foregroundServiceType attribute specified [ForegroundServiceType]
        startForeground(notificationId, notification)
        ~~~~~~~~~~~~~~~

   Explanation for issues of type "ForegroundServiceType":
   For targetSdkVersion >= 34, to call Service.startForeground(), the
   <service> element in the manifest file must have the foregroundServiceType
   attribute specified.


https://github.com/android/snippets/actions/runs/35001625484/job/104492132218?pr=1079

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
```diff
@@ -112,6 +112,7 @@ private class TimeSensitiveNotificationActivity : ComponentActivity() {
 private class TimeSensitiveOngoingService : Service() {
     override fun onBind(intent: Intent?): IBinder? = null
 
+    @SuppressLint("ForegroundServiceType")
```
alabiaga: It's a snippet and doesn't need to run but the issue is that the code will compile and i need to verify but willl crash at runtime if it targets Android 14 (API level 34) or higher.

Either declare a fgs type here and in manifest or comment here that we suppress the type but it should defined for a production app.

If doing the former, we can use any type that doesn't require a runtime permission request:

e.g. https://developer.android.com/develop/background-work/services/fgs/service-types#short-service

hamen: Declared the type instead of suppressing it. `compose/snippets/src/main/AndroidManifest.xml` now carries `<service android:name=".notifications.TimeSensitiveOngoingService" android:exported="false" android:foregroundServiceType="shortService" />`, together with the `FOREGROUND_SERVICE` and `FOREGROUND_SERVICE_SHORT_SERVICE` permissions. `shortService` needs no runtime permission request, as you suggested.

The `@SuppressLint` is gone. The class is no longer private, because the framework instantiates it.

The region tag content does not change, so the snippet on the page stays the bare `startForeground()` call.

hamen:
@kkuan2011 the `ForegroundServiceType` build error is fixed in d9e66b18. The service declares `android:foregroundServiceType="shortService"` in the manifest, which is what Aaron asked for in his review, so the lint check has nothing left to report and the code no longer throws on API 34 and higher. The branch is also up to date with `main` again.

kkuan2011 (approved):
Approved, assuming the build passes (Still running)

alabiaga (approved):

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
```diff
@@ -0,0 +1,124 @@
...
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.core.app.NotificationCompat
+import com.example.compose.snippets.R
+
+private const val CHANNEL_ID = "channelId"
+
+// [START android_notifications_time_sensitive_channel]
+class DACapp : Application() {
```
riggaroo: Can this be named more appropriately?

riggaroo (changes requested):

compose/snippets/src/main/java/com/example/compose/snippets/notifications/TimeSensitive.kt
hamen: Done in 8545c28. It is now `MyApplication`, the name the Android documentation uses for an `Application` subclass elsewhere, and the one that matches `SampleStackTracesEnabledApp` in this module.

The class sits inside the region tag, so the page carries `DACapp` until the page text gets the same rename. Say the word if you prefer a different name.

riggaroo (approved):

PR #1082, opened 2026-09-10
Migrate Splash screens snippets

hamen:
Code snippets are for:
- https://developer.android.com/develop/ui/views/launch/splash-screen

List of modifications:
- Splash screens page: Added Kotlin code snippets for suspend drawing with OnPreDrawListener, exit animation listener, and remaining duration calculation.
- D22: Preserved // ... placeholder comment with exact column indentation, placing compiling setup code inside [START_EXCLUDE silent].
- D25: Reused existing views module without modifying settings.gradle.kts.

Snippets not migrated:
- Lines 235-249: Gradle dependency block
- Lines 262-264: Theme attribute XML snippet
- Lines 279-281: Theme attribute XML snippet
- Lines 294-296: Theme attribute XML snippet
- Lines 302-304: Theme attribute XML snippet
- Lines 311-313: Theme attribute XML snippet
- Lines 325-327: Theme attribute XML snippet

### DIFF
--- a/views/src/main/java/com/example/example/snippet/views/launch/SplashScreen.kt
+++ b/views/src/main/java/com/example/example/snippet/views/launch/SplashScreen.kt
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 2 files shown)


views/src/main/java/com/example/example/snippet/views/launch/SplashScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+        )
+    }
+    // [END android_views_splash_screen_suspend_drawing]
+}
+
+class ExitAnimationActivity : Activity() {
+    // [START android_views_splash_screen_exit_animation]
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        // [START_EXCLUDE silent]
```
kkuan2011, line 70: nit: you could remove silent so it's just [START_EXCLUDE]

and then remove line 73

kkuan2011 (approved):
Looks good, assuming we want to migrate these views snippets from Aaron's perspective

views/src/main/java/com/example/example/snippet/views/launch/SplashScreen.kt
hamen: Done ✅ Dropped `silent` and the hand-written `// ...`, so the exclude prints the placeholder itself.

PR #933, opened 2026-05-29
Add remaning notification snippets for d.android.com

alabiaga:
Change-Id: I5a9d083affc9ec7b9be5542a23933f694450a2b3

Snippets for
- https://developer.android.com/develop/ui/compose/notifications/channels
- https://developer.android.com/develop/ui/compose/notifications/bubbles
- https://developer.android.com/develop/ui/compose/notifications/call-style

PR #1084, opened 2026-09-11
Migrate App shortcuts snippets

hamen:
Code snippets are for:
- https://developer.android.com/develop/ui/compose/system/shortcuts/creating-shortcuts
- https://developer.android.com/develop/ui/compose/system/shortcuts/managing-shortcuts
- https://developer.android.com/develop/ui/compose/system/shortcuts/adding-capabilities

Region tags: `android_shortcuts_xml_static`, `android_shortcuts_create_dynamic`, `android_shortcuts_pin_shortcut` (Create shortcuts), `android_shortcuts_restore_dynamic` (Manage shortcuts), `android_shortcuts_xml_capability` (Add capabilities).

List of modifications:
- Create shortcuts page, `android_shortcuts_pin_shortcut`: D3. The page passes `/* flags */ 0` to `PendingIntent.getBroadcast()`. A mutability flag has been mandatory since API level 31, so the published call throws on any current device. It reads `PendingIntent.FLAG_IMMUTABLE` here. Nothing in the block calls `setPendingIntentTemplate()` or `setOnClickFillInIntent()`, which is what would make it `FLAG_MUTABLE`.
- Manage shortcuts page, `android_shortcuts_restore_dynamic`: D24. The page calls `ShortcutManagerCompat.getPinnedShortcuts(this)`, and that method does not exist. Typing the page's own token into the module gives `Unresolved reference 'getPinnedShortcuts'`; `androidx.core` publishes `getDynamicShortcuts()` and `getShortcuts(Context, int)` and has never published `getPinnedShortcuts` at any version. The block reads `ShortcutManagerCompat.getShortcuts(this, ShortcutManagerCompat.FLAG_MATCH_PINNED)`, written out in full rather than through an import, because an import sits outside the region and would leave the published block showing a bare constant with no owner. The page's block does not compile as published, so this wants a page edit as well.
- Create shortcuts page, `android_shortcuts_pin_shortcut`: the block now uses the AndroidX Compat API, on review. `ShortcutManagerCompat.isRequestPinShortcutSupported()`, `ShortcutInfoCompat.Builder`, `ShortcutManagerCompat.createShortcutResultIntent()` and `ShortcutManagerCompat.requestPinShortcut()` replace the platform `ShortcutManager` and `ShortcutInfo`. The other two Kotlin blocks on these pages already used the Compat classes, so this makes all three agree. The page prints the platform version, so this wants a page edit.
  This also settles a second defect in the same block. The page named one thing two ways, a bare `getSystemService()` that needs a `Context` receiver, and a `context` variable two lines down. The Compat API is static and takes the `Context` as an argument, so the `getSystemService()` line and its non-null assertion are gone rather than repaired.
  One caveat for the page owner: `ShortcutInfoCompat.Builder(context, "my-shortcut").build()` throws without a short label and an intent, where the platform builder does not. The page prints a bare `.build()`, so that line now fails at runtime instead of only being incomplete. The snippet keeps the page's shape. Say the word if you want it to set a label and an intent.
- D19: `spotlessApply` moved four lines in `android_shortcuts_create_dynamic` and four in `android_shortcuts_pin_shortcut`: four-space continuation indents, one argument per line, and no blank line before a closing brace. `android_shortcuts_restore_dynamic` moved one. The two XML blocks are verbatim, 20/20 and 12/12.
- D8: `R.drawable.icon_website` is scaffolding written for this extraction, because `android_shortcuts_create_dynamic` passes it to `IconCompat.createWithResource()` and it has to resolve. It is the Material `language` icon; `compose_icon`, named by the static shortcuts XML, is the Material `widgets` icon. Both record their source in the file. Neither declares a tint: the guide's own attribute table says a shortcut icon cannot carry one, and the launcher draws them outside this module's theme.
- The two XML blocks live in `compose/snippets/src/main/res/xml/`, and the four resources the static block names are declared in the module.
- The three shortcut label strings the static XML block names are declared in `values/strings.xml`, and in `values-es/strings.xml` as well. Without the Spanish entries `lintDebug` fails with `MissingTranslation`.

Snippets not migrated:
- Create shortcuts, line 33: `AndroidManifest.xml` activity with the shortcuts `<meta-data>`. Manifest fragment; merging one changes the application manifest.
- Create shortcuts, line 242: `gradle.properties` for AndroidX. Build configuration, not module code.
- Create shortcuts, line 249: `dependencies { }` for the Google Shortcuts Integration Library. Build script, and it pins versions that would go stale here.
- Create shortcuts: ten indented spans under *Customize attribute values* and *Configure inner elements*. Not code. They are the descriptions of a definition list, which the Markdown export indents exactly as it indents a code block; line 87 is the sentence *"A string literal that represents the shortcut when a `ShortcutManager` object performs operations on it."*
- Manage shortcuts, line 228: `adb shell cmd shortcut reset-throttling`. A shell command.

No Java blocks on any of the three pages.

D21, two sections that promise code and show none:
- Create shortcuts, *Create a custom shortcut activity*. The reader is told to add `ACTION_CREATE_SHORTCUT` to an activity's `<intent-filter>`, then given five steps naming `createShortcutResultIntent()`, `setResult()` and `finish()`. Only `createShortcutResultIntent()` appears in any block on the page, and that block is about pinning.
- Manage shortcuts, *Start one activity from another*. Two numbered steps: put `android:taskAffinity=""` on the trampoline activity in `AndroidManifest.xml`, and reference that activity from the intent in the shortcuts resource file. No block for either.

Writing those blocks here would be new teaching content rather than a copy of anything. Every other step on all three pages was read against the blocks and is answered by one.


compose/snippets/src/main/java/com/example/compose/snippets/system/ShortcutsSnippets.kt
```diff
@@ -0,0 +1,94 @@
...
+        )
+        .build()
+
+    ShortcutManagerCompat.pushDynamicShortcut(context, shortcut)
+    // [END android_shortcuts_create_dynamic]
+}
+
+fun pinShortcut(context: Context) {
+    // [START android_shortcuts_pin_shortcut]
+    val shortcutManager = context.getSystemService<ShortcutManager>()
```
alabiaga, line 53: Should've been fixed long ago but lets use ShortcutInfoCompat and the relevant methods when used below.

https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat

Otherwise looks good to me.

alabiaga (changes requested):

compose/snippets/src/main/java/com/example/compose/snippets/system/ShortcutsSnippets.kt
hamen: Done, thanks. The pin_shortcut block was the only one still on the platform API. The other two already used ShortcutManagerCompat.

`isRequestPinShortcutSupported`, `createShortcutResultIntent` and `requestPinShortcut` now all go through `ShortcutManagerCompat`, and the shortcut is built with `ShortcutInfoCompat.Builder`. This removes the `getSystemService<ShortcutManager>()` line and its `!!` as a side effect.

This is a page change too. The guide still prints the platform version, so I added it to the page-edit list in the description.

The same push fixes the `lintDebug` failure. This change adds three shortcut label strings, and they had no entry in `values-es`, so lint reported `MissingTranslation`. I added the Spanish entries, the way #933 handled a new string.

One thing to flag: `ShortcutInfoCompat.Builder(...).build()` throws without a short label and an intent, where the platform builder does not. The page prints a bare `.build()`, so the published line now fails at runtime instead of only being incomplete. Tell me if you want the snippet to set a label and an intent. That is a larger difference from the page.

alabiaga: >  Tell me if you want the snippet to set a label and an intent. That is a larger difference from the page.

Yes, let's add those requirements.

alabiaga (approved):

alabiaga:
@riggaroo can you help review/unblock for submission. Note that like notifications, shortcuts are ui tooling agnostic, so very little to no compose related code here.

riggaroo (approved):

PR #1085, opened 2026-09-11
Add a navDeepLink routing snippet for Receiving simple data

hamen:
Code snippets are for:
- https://developer.android.com/develop/ui/compose/sharing/receive

**This lands with a `receive.md` edit, and the two have to ship together.** The page edit is already staged and ships with this PR. Two of the three regions below no longer match the live page, on purpose, because the page is changing to match them.

The routing snippet was written at the page owners' request. A written brief went to them naming three findings on this page: the prose contradicting its only snippet, the two-step process being incomplete because `navDeepLink` routes but does not deliver `EXTRA_STREAM` or `EXTRA_TEXT`, and two sentences of View system text with no snippet behind them. They were offered the code following the prose or the prose following the code, and chose the first. The guide told the reader to route the incoming `ACTION_SEND` intent with Compose Navigation's deep link support, *"instead of manually intercepting and parsing the Intent"*, and no `navDeepLink` code appeared anywhere on the page — the one snippet did exactly the manual interception the prose says to avoid.

List of modifications:
- `android_receive_data_nav_deeplink`: new region, the missing routing step. Tagged around the `composable(route = "incoming_share", ...)` call alone, with no exclude blocks, so the page publishes that call and nothing else. It declares three links: `ACTION_SEND` with `text/plain`, `ACTION_SEND` with `image/*`, and `ACTION_SEND_MULTIPLE` with `image/*`. The third is there because the manifest declares that filter and `SharesheetHandler` handles it, so leaving it out would give a reader who follows steps 1 and 2 no route for multiple images.
- `android_handle_intent_action_data_sent`: edited, and no longer verbatim against the live page. It carries the rewrite that ships with the `receive.md` edit — `private` handlers, `Intent.ACTION_SEND` in place of the bare imported `ACTION_SEND`, `?.let` on the nullable image extras, and reworded comments. 46 lines here against the page's 43. `spotlessApply` moved nothing.
- `android_handle_intent_handle_extra_text`: edited, one comment. It was 6/6 verbatim and is now 5/6, because the reworded comment collapses the page's two comment lines into one. That wording ships in `receive.md` with the rest.

Snippets not migrated:
- The `AndroidManifest.xml` activity block with the `ACTION_SEND` intent filters. Manifest snippets are not migrated, so it stays hardcoded on the page.

One thing for the page while that manifest block is being edited: the activity declares intent filters and no `android:exported`, so as printed today it will not install on API level 31 or higher. It wants `android:exported="true"` to receive incoming share intents. That is a defect on the live page rather than a nicety.

Note on overlap: this touches the same file as #1063, which removes the `ChooserTarget` import and reworks `customTargets()`. The two overlap only in the import block; whichever lands second takes a one-line conflict there.


compose/snippets/src/main/java/com/example/compose/snippets/sharesheet/SharesheetSnippets.kt
```diff
@@ -29,18 +29,53 @@ import android.service.chooser.ChooserTarget
...
+        }
+        // [END android_receive_data_nav_deeplink]
+    }
+}
+
 // [START android_handle_intent_action_data_sent]
 @Composable
 fun SharesheetHandler() {
     val context = LocalContext.current
     val intent = (context as? Activity)?.intent
```
bsagmoe, line 73: I think you could probably update this to use `LocalActivity`:

```kotlin
val activity = LocalActivity.current
val intent = activity?.intent
```

hamen: Done in fc30bf1, thanks! Switched to `LocalActivity.current` and dropped the now unused `android.app.Activity` and `androidx.compose.ui.platform.LocalContext` imports, since this was their only usage in the file. `spotlessCheck` passes locally.

compose/snippets/src/main/java/com/example/compose/snippets/sharesheet/SharesheetSnippets.kt
```diff
@@ -81,8 +118,7 @@ fun handleSendAndExtraText(intent: Intent) {
     IntentCompat.getParcelableExtra(intent, Intent.EXTRA_STREAM, Uri::class.java).let {
```
kkuan2011: Should we do the same here as above?

val imageUri = IntentCompat.getParcelableExtra(intent, Intent.EXTRA_STREAM, Uri::class.java)
    imageUri?.let {
       ...

bsagmoe (approved):

PR #1092, opened 2026-09-14
Added coroutinescope to avoid any complication errors in skill

niharika2810:


kkuan2011 (commented):
Thanks Niharika! Could an identity team member approve first? 

It looks like the end tag is in the wrong place (and we would have un-closed brace in the snippet). 

Should it be coroutineScope.launch { ... } instead of coroutineScope { } so that the function can be called from a non-suspending context? Or should the suspend function signature be shown inside the region tags?

niharika2810:
> Thanks Niharika! Could an identity team member approve first?
> 
> It looks like the end tag is in the wrong place (and we would have un-closed brace in the snippet).
> 
> Should it be coroutineScope.launch { ... } instead of coroutineScope { } so that the function can be called from a non-suspending context? Or should the suspend function signature be shown inside the region tags?

Hey Chiping is OOO and I cant approve my own PR. Maybe if you can Joe or Vinisha? 

coroutineScope - This is only for demo purpose - just to make sure apps know that this is a suspend function. So should be good. 

I checked it in Android Studio - braces look fine to me - do you see anything here?

kkuan2011:
Discussed the feedback with Niharika and will proceed with merging (Chiping is ooo).

PR #1093, opened 2026-09-14
Migrate Create custom haptic effects snippets

hamen:
Code snippets are for:
- https://developer.android.com/develop/ui/views/haptics/custom-haptic-effects

List of modifications:
- Custom haptic effects page: Added 16 Kotlin snippets for custom vibration patterns, vibration compositions (Resist, Expand, Wobble, Bounce), and waveform envelopes (BasicEnvelopeBuilder, WaveformEnvelopeBuilder, BouncingSpringAnimation, RocketLaunchAnimation, LavaBeats).
- Custom haptic effects page: Repaired compilation and syntax issues (Java-in-Kotlin void/type declarations in repeating pattern, variable name and spread operator in arePrimitivesSupported, missing closing brace in ExpandScreen clickable lambda, rampUpDurationMs identifier and Long type compatibility in RocketLaunchAnimation).
- Custom haptic effects page: Added background color resource in views/src/main/res/values/colors.xml and enabled Compose support in views/build.gradle.kts.
- Declared android.permission.VIBRATE in the views manifest, which every vibrate() call needs.

Deviations from the page:
- BouncingSpringAnimation calls `DrawSpring(springX, springY)`. The page calls `DrawSpring(mutableStateOf(springX), mutableStateOf(springY))`, which creates two state objects on every recomposition. The page text needs the same change.
- RocketLaunchAnimation reads the background with `colorResource(R.color.background)`. The page uses `Color(context.getColor(R.color.background))`, which is not configuration-aware and can return a stale value after a configuration change. The `LocalContext.current` lookup goes with it.
- `arePrimitivesSupported()` takes a vararg, so the Kotlin snippet passes `*primitives`. The page omits the spread operator, and that version does not compile.

Snippets not migrated:
- 11 duplicate Java snippets, since the page is served Kotlin-first.


views/build.gradle.kts
```diff
@@ -47,9 +48,18 @@ android {
             jvmTarget = JvmTarget.fromTarget("17")
         }
     }
+    buildFeatures {
+        compose = true
+    }
 }
 
 dependencies {
+    val composeBom = platform(libs.androidx.compose.bom)
```
kkuan2011, line 57: how come we need to add compose dependencies in the view module? should we just add those snippets in the compose module then?

hamen: Done ✅ Moved to `compose/snippets/src/main/java/com/example/compose/snippets/haptics/CustomHapticEffects.kt`. `views` is back to main, and the tags are now `android_haptics_*`, so the page includes need the new names.

compose/snippets/src/main/java/com/example/compose/snippets/haptics/CustomHapticEffects.kt
```diff
@@ -0,0 +1,800 @@
...
+        vibrator.vibrate(
+            VibrationEffect.createWaveform(
+                onOffTimings, onOffRepeatIdx
+            )
+        )
+    }
+    // [END android_haptics_custom_fallback]
+}
+
+@RequiresApi(Build.VERSION_CODES.R)
```
kkuan2011, line 130: I believe all @RequiresApi annotations are not needed in this file, and remove build import too

kkuan2011 (approved):
Let's wait for Aaron to return to review since he's the closest to haptics

compose/snippets/src/main/java/com/example/compose/snippets/haptics/CustomHapticEffects.kt
hamen: Done! Removed all `@RequiresApi` annotations, `import android.os.Build`, and `import androidx.annotation.RequiresApi` in `a1c2566d`.

PR #1112, opened 2026-09-17
Migrate Snippet: App fundamentals

pengzhi-cb:
Here are the code snippets per URL that need to be migrated:

1. (3 snippets) https://developer.android.com/guide/components/activities/intro-activities
1. (5 snippets) https://developer.android.com/guide/components/activities/recents
1. (5 snippets) https://developer.android.com/guide/components/activities/secure-bal
1. (6 snippets) https://developer.android.com/guide/topics/resources/app-languages
1. (1 snippets) https://developer.android.com/guide/topics/resources/localization
1. (8 snippets) https://developer.android.com/guide/topics/resources/string-resource

---

**Total: 28 code snippets across 6 pages.**


compose/snippets/src/main/java/com/example/compose/snippets/activities/IntroActivities.kt
```diff
@@ -0,0 +1,102 @@
...
+
+      else -> {
+          // Handle other intents
+      }
+    }
+  }
+}
+// [END android_activities_intro_example_activity]
+
+class MyViewModel : ViewModel() {
```
kkuan2011, line 95: mark as private class

compose/snippets/src/main/java/com/example/compose/snippets/resources/StringResource.kt
```diff
@@ -0,0 +1,164 @@
...
+    val text = buildAnnotatedString {
+        append(source.toString())
+        source.getSpans(0, source.length, Annotation::class.java)
+            .forEach { annotation ->
+                if (annotation.key == "font" &&
+                    annotation.value == "title_emphasis") {
+                    addStyle(
+                        SpanStyle(
+                            fontFamily = FontFamily(
+                                Font(R.font.permanent_marker)
```
kkuan2011, line 153: Instead of adding a new asset to the snippets repo,  can we use an existing font in the snippets repo? like R.font.raleway_regular (and remove permanent_marker.ttf) It's okay if it makes a visible difference in the code on the DAC page.

compose/snippets/src/main/java/com/example/compose/snippets/activities/SecureBal.kt
```diff
@@ -0,0 +1,124 @@
...
+import android.os.StrictMode
+import android.util.Log
+import androidx.activity.ComponentActivity
+import androidx.annotation.RequiresApi
+import java.util.concurrent.Executor
+
+private const val TAG = "SecureBal"
+private const val REQUEST_CODE = 100
+
+@RequiresApi(Build.VERSION_CODES.UPSIDE_DOWN_CAKE)
```
kkuan2011, line 35: I think you have some unnecessary @RequiresApi annotations throughout the files, can you double check if they're still needed? I believe the minSDK on the compose module is 36.

Note: If there are @RequiresApi annotations within the region tags (meaning they show up in the visible code on DAC), then you can leave those in the code.

kkuan2011 (commented):
Also run `./gradlew :compose:snippets:spotlessApply` to normalize 2-space vs. 4-space indentation in **`compose/snippets/src/main/java/com/example/compose/snippets/activities/IntroActivities.kt` (Lines 45–95)** & **`AppLanguages.kt` (Lines 39–41, 126–133)**:

PR #1114, opened 2026-09-17
Media2 snippet migration

barbaralaw:
## Summary

Extracts the Kotlin samples from eight Media guides (AI enhancement, HDR, media projection, spatial audio, and Ultra HDR) into `:media` as region-tagged source, so the guides can import them instead of hardcoding them.

**19 snippets across 8 pages. 14 publish exactly what the page shows. 5 differ, listed below with a reason.** Small Spotless rewraps (a long call split the same way the formatter always splits it) are not listed. A wrap that changes the shape of the call is listed.

Region tags all begin `android_media_`; the tables drop that prefix.

### How to read the "why" column

| Why | What a reader of the page would see |
|---|---|
| **hardcoded snippet defect fixed** | The published Kotlin does not compile, and this is not a sample meant to demonstrate an error. The region publishes the corrected line. |
| **`// ...`** | The page prints a bare `...`, which Kotlin will not compile. The region publishes `// ...`. |
| **comment indent** | Comments that sat at column 0 are indented with the surrounding class or method body so DevSite can strip common indent. |
| **Spotless wrap** | Same tokens, different line breaks. ktlint moved the call enough that the include does not look like the fenced block. |

### 1. Understand the media enhancement lifecycle in Bitmap mode

https://developer.android.com/media/ai-enhancement/bitmap-mode-lifecycle

3 snippets, 1 match, 2 differ.

| Snippet | Page section | Difference | Why |
|---|---|---|---|
| `ai_enhancement_bitmap_wrappers` | Create session and bitmap process wrappers | `this.createSession(...)` inside `withContext { }` becomes `this@createSessionAsync.createSession(...)` | hardcoded snippet defect fixed |
| `ai_enhancement_bitmap_viewmodel` | Execute the bitmap pipeline in a ViewModel | Comments that were at column 0 are indented with the class/method body | comment indent |

Matching as published: `ai_enhancement_bitmap_initialize_engine`.

The enhancement client types used by this and the Bitmap/Surface/get-started pages are compile-only shims in `:media`. The get-started page’s Gradle coordinate (`play-services-media-effect-enhancement`) was left hardcoded.

### 2. Get started with Media Enhancement APIs

https://developer.android.com/media/ai-enhancement/get-started

1 snippet, matches as published: `ai_enhancement_get_started_checks`.

### 3. Understand the media enhancement lifecycle in Surface mode

https://developer.android.com/media/ai-enhancement/surface-mode-lifecycle

2 snippets, all 2 match as published: `ai_enhancement_surface_initialize_engine`, `ai_enhancement_surface_snapshot`.

### 4. Color correct with look-up tables (LUTs)

https://developer.android.com/media/grow/hdr-lut

1 snippet, matches as published: `hdr_lut_apply`.

### 5. HDR video playback

https://developer.android.com/media/grow/hdr-playback

2 snippets, 1 match, 1 differ.

| Snippet | Page section | Difference | Why |
|---|---|---|---|
| `hdr_playback_mediacodec` | Kotlin | Split a jammed `val list = MediaCodecList(...) var format = MediaFormat() …` onto two statements | hardcoded snippet defect fixed |

Matching as published: `hdr_playback_check_support`.

### 6. Media projection

https://developer.android.com/media/grow/media-projection

3 snippets, 2 match, 1 differ.

| Snippet | Page section | Difference | Why |
|---|---|---|---|
| `projection_window_context_metrics` | Resizable apps | `createWindowContext(context.display!!, TYPE_APPLICATION, null)` is split so `context.display!!` and the closing `)` sit on their own lines | Spotless wrap |

Matching as published: `projection_start`, `projection_virtual_display`.

### 7. Spatial Audio

https://developer.android.com/media/grow/spatial-audio

6 snippets, 5 match, 1 differ.

| Snippet | Page section | Difference | Why |
|---|---|---|---|
| `spatial_audio_disable_channel_constraints_selector` | Disable device-capability constraints | Bare `...` becomes `// ...` | `// ...` |

Matching as published: `spatial_audio_get_spatializer`, `spatial_audio_disable_channel_constraints_player`, `spatial_audio_set_max_channels`, `spatial_audio_max_output_channels`, `spatial_audio_audio_format`.

### 8. Display Ultra HDR images

https://developer.android.com/media/grow/ultra-hdr/display

1 snippet, matches as published: `ultra_hdr_display_window_color_mode`.

## Not extracted

Kotlin only. HDR LUT, HDR playback, media projection, spatial audio, and Ultra HDR each have Java twins left for the page edit that retires them. Get started also leaves two Gradle `dependencies` blocks and the native-library manifest. Media projection leaves the foreground-service manifest. Spatial audio leaves the dumpsys/log “Test spatial audio” output.

## Live snippet fixes this surfaced

1. **HDR video playback:** one Kotlin sample joins two declarations on a single line (`val list = MediaCodecList(...) var format = MediaFormat() /* media format from the container */;`), which is not valid Kotlin. The region publishes them as separate statements.
2. **Bitmap-mode lifecycle wrappers:** `this.createSession(...)` sits inside `withContext(Dispatchers.Main) { }`, so `this` is the `CoroutineScope` and the call does not compile. The region keeps the Main hop (the ViewModel sample calls this wrapper from `Dispatchers.IO`) and publishes `this@createSessionAsync.createSession(...)`.
3. **Spatial audio:** a standalone `...` between `DefaultTrackSelector` setup and `buildUponParameters` is not valid Kotlin. The region publishes `// ...`.

## Verification

- `./gradlew :media:compileDebugKotlin` — passes
- `./gradlew :media:lintDebug` — passes
- `./gradlew :media:spotlessApply` — run; Kotlin in `:media` is already clean. Spotless skipped `media/build.gradle.kts` (formatter does not converge on that file).


media/src/main/java/com/example/media/BitmapModeLifecycle.kt
```diff
@@ -0,0 +1,256 @@
...
+        // 4. Critical: Release native GPU hardware resources
+        enhancementSession?.release()
+        enhancementSession = null
+        enhancementExecutor.shutdown()
+        super.onCleared()
+    }
+}
+// [END android_media_ai_enhancement_bitmap_viewmodel]
+
+// Shims for Media Enhancement API types if not provided by standalone SDK
```
kkuan2011, line 201: Are these classes provided by an existing google library? If so, could we use the library instead of defining shims?

media/src/main/java/com/example/media/UltraHdrDisplay.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.example.media
+
+import android.content.pm.ActivityInfo
+import android.graphics.Bitmap
+import android.widget.ImageView
+import androidx.fragment.app.Fragment
+
+class UltraHdrDisplayFragment : Fragment() {
```
kkuan2011, line 24: If this is more of a Fragment snippet, I'm wondering if we need to migrate this? Is there a compose-version of this code? or is this the only way to do it?

barbaralaw: Thanks, these are good questions. The published Kotlin is the Views path: `ImageView` via view binding and `requireActivity().window.colorMode`. The Fragment wrapper is only there because `requireActivity()` is a Fragment API. 

Looking into it more though, on the [Overview page](https://developer.android.com/media/grow/ultra-hdr) for this section there is a link to github resource on [Displaying an Ultra HDR image in Compose](https://github.com/android/platform-samples/pull/89). It seems to me that this section of the page needs work beyond the scope of the migration and should be added to the list of other snippets that need further work, leaving this snippet hardcoded for now. Does that sound OK?

